PR #2, opened 2021-02-01
Manjaro: Remove generic part from the wifi install and reference guides/wifi

networkException:
re: @JPyke3

What do you think of this change?


JPyke3:
Perfect! I think that's a much better way to have it instead of duplicating content 👍

PR #5, opened 2021-02-11
Put the dkms and windows guides in  mkdocs.yml

Redecorating:
There's no link to them right now because I forgot. Again, ~~first try~~


networkException:
Didn't think about it either, thank you :+1:

Issue #9, opened 2021-02-12
Adding Pacman Key missing in arch Install Guide


Redecorating:
There's no instruction to add aunali1's package signing key to new Arch installs, in the Arch Install Guide. This conversation has how to do it https://canary.discord.com/channels/595304521857630254/722427464604844122/770713657012650024 .

PR #10, opened 2021-02-12
Arch: Added key importing instructions to the install guide

networkException:
See #9﻿


networkException:
Do you think this is good @Redecorating?

Redecorating:
That looks good

PR #18, opened 2021-02-23
windows: a command needed changing

Redecorating:
It assumed you were in a particular directory


networkException:
Thank you for the fix

PR #29, opened 2021-05-12
Add NixOS documentation

mikroskeem:


docs/distributions/nixos/installation.md
```diff
@@ -0,0 +1,5 @@
+# Installing NixOS
+
+_tbd_
+
+If you have issues, feel free to ask on our [Discord Server](https://discord.gg/Jayz5f5).
```
networkException, line 5: I'd say with the announcement bar we don't really have to link the server everywhere (see https://github.com/t2linux/wiki/commit/063fb41d0f3edd4f9afdc5020ab4f62903763262). What do you think?

mikroskeem: I think having it on announcement bar is indeed better. I'll drop this as I'll start adding more content

mikroskeem:
Will be superseded by #315

PR #32, opened 2021-05-23
Kernel compile/install guide

### LINKED ISSUE #26: Guides: Write dedicated compile guide
Currently distributions/ubuntu/building is used as the patching and compiling a kernel guide. Besides that not being easy to discover by a user, the article also makes a lot of assumptions and gives instructions without much explentation of the steps.

Redecorating:
Add instructions for compiling linux-mbp-arch on non arch based systems, and instructions for extracting the compiled kernel, modules, and related files from the compiled linux-mbp-arch package. Should resolve #26 

In guides/kernel.md, we might want to just make people build a package instead of installing things directly to their filesystem, as it's better (in my opinion) to use a package manager for this. That would make the page layout better. However, this would be a problem if they can't build a package (some package managers might be very complicated). I'll look into changing that tomorrow if it's wanted.


networkException:
Alright just did some cleanup and removed duplicated stuff. I also removed the arch makepkg instructions, I think it would be the best to also have them in guides/kernel

PR #56, opened 2021-07-13
wifi: model, chip, fw table

Redecorating:
this is a draft, if ppl have missing bits, tell me or add them

PR #75, opened 2021-07-21
Update wifi.md

AdityaGarg8:


networkException:
I merged this for now as the information is not strictly correct anymore, however because users commonly missed to use the correct firmware in the past #56 should probably reintroduce it

AdityaGarg8:
> I merged this for now as the information is not strictly correct anymore, however because users commonly missed to use the correct firmware in the past #56 should probably reintroduce it

```
Note: If you are unable to get the firmware even after following guide for other models, you might need to extract firmware from macOS at this point. See the instructions in On any OS section, especially the first note there.
```

I have already put this in the wifi guide

AdityaGarg8:
I must say Redecorating's guide is much much better

PR #83, opened 2021-07-26
Update roadmap.md

AdityaGarg8:


docs/roadmap.md
```diff
@@ -39,9 +39,7 @@ If you wish to use another distribution, you can install it normally, also follo
 
 ## Preparing the Installation
 
-You will want to look at [Is my model supported?](https://wiki.t2linux.org/guides/wifi/#is-my-model-supported) in the wifi guide before starting an installation.
-Based on that information, follow the wifi guide until any steps you need to run on linux. This includes finding
-out your model identifier, the firmware files needed for your device and downloading or extracting those firmware files.
+It is recommended follow the [wifi guide](https://wiki.t2linux.org/guides/wifi/) until any steps you need to run on linux. This includes finding out your model identifier, the firmware files needed for your device and downloading or extracting those firmware files. In case you need to get another kernel as mentioned in the guide, get that too. It is also recommended to go through the [DKMS](https://wiki.t2linux.org/guides/dkms/) guide if using a new kernel as you may have to uninstall old drivers.
```
networkException, line 42: I don't really know if the dkms guide should be mentioned here already, wouldn't that at least require the new kernel to be installed?

AdityaGarg8: A new kernel has to be installed to get wifi working in some cases.

AdityaGarg8: Before that old modules need to be removed and new installed (In Ubuntu atleast). Else the driver breaks

networkException: I don't think such instructions belong in the roadmap. The wifi guide should contain all information to get a patched kernel fully up and running if thats needed, including dkms. If we keep having partial instructions for a topic and only mention the full instructions in some other guide the wiki will get confusing fast

AdityaGarg8: Alright then. You may close the request. Just do some editing to comply with the new wifi guide.

networkException: I will update other pages accordingly then. Can you just remove this change? I will merge the other grammar fixes then

PR #94, opened 2021-07-28
Add refind guide

AdityaGarg8:


AdityaGarg8:
Its not completed yet. Pull request put for suggestions.

AdityaGarg8:
Shall write further, later.

networkException:
Please convert it to a draft in that case

docs/guides/refind.md
```diff
@@ -0,0 +1,47 @@
...
+
+All steps given here have to be performed on **macOS**. You will also need to have [secure boot disabled](https://support.apple.com/en-us/HT208198).
+
+1. With the help of disk utility, create a 100-200MB `MS-DOS FAT` partition and label it as `REFIND`.
+2. Get a **binary zip file** of rEFInd from [here](https://www.rodsbooks.com/refind/getting.html).
+3. The binary zip file of rEFInd shall be available in the downloads folder by the name of `refind-bin-<VERSION>.zip`, where `<VERSION>` represents the version of refind you have downloaded. For eg:- If you have downloaded `0.13.2` version, it will be available as `refind-bin-0.13.2.zip`.
+4. Extract the zip file (can be done by double clicking on it). The contents shall be extracted in a folder named `refind-bin-<VERSION>`. Here `<VERSION>` means the same as described in step 3.
+5. Open the terminal and run `diskutil list` to get the disk identifier of the `REFIND` volume created in step 1. A sample output is given below:-
+
+   ```sh
```
Redecorating, line 15: You might want to use `plain` as the language here, as there's a `*` that gets highlighted. Using `sh` does make the heading lines grey, which is nice for readability.

AdityaGarg8: Thanks

AdityaGarg8:
Complete from my side. You may test it out and review it.

AdityaGarg8:
I guess this guide should get rEFInd running smoothly. The only issue which may be faced is with the iGPU, as rEFInd boots `/EFI/Ubuntu/Grubx64.efi` instead of `/EFI/BOOT/Bootx64.efi`. I guess a workaround can be replacing grubx64.efi instead of bootx64.efi with the compiled EFI file. Though rEFInd also has a spoof feature which can be an easier approach. The only this is that I haven't tested all this and won't be able to as my Medical Entrance Test is very close and thus I have to study now. Though I shall be maintaining the Ubuntu kernels as I am now, as it is a hardly 1 min job. So, I would rather say that someone else may test the guide.

AdityaGarg8:
Also do inform me if you encounter any issue in this guide. I may have missed something out.

networkException:
LGTM so far 👍

@Redecorating could you take a look at this as well? If you say it's fine I will merge.

Redecorating:
It looks good

PR #120, opened 2021-08-12
Move wifi to working section.

AdityaGarg8:


networkException:
This is fine even though some models have unknown firmware still, we can just change it to exclude some specific model if we find out about incompatible systems in the future

AdityaGarg8:
> This is fine even though some models have unknown firmware still, we can just change it to exclude some specific model if we find out about incompatible systems in the future

Yeah

PR #124, opened 2021-08-13
Create keyboard.md

dkuku:
how to swap modifier keys


networkException:
I don't think copying the archwiki is a good approach here as we don't have anything t2 specific to add really. Please just add a link to the archwiki page in index.md.

PR #126, opened 2021-08-13
Update index.md

dkuku:
add link to arch wiki


networkException:
For the record:

See #124

PR #141, opened 2021-08-28
Initial attempt to fix numbering

AdityaGarg8:


AdityaGarg8:
Preview on GitHub shows correct numbering but on wiki after step 6 it changes to 1.

AdityaGarg8:
I guess the only way out to check this is to merge this

Redecorating:
I think the issue may be that  you are using 3 spaces to indent the things in each dot point. In the hybrid graphics page, the numbering is fine and I think that's because it uses 4 spaces,

AdityaGarg8:
Let me do it to 4

AdityaGarg8:
It's fixed now

PR #151, opened 2021-09-14
Added back alternate Ubuntu repo

AdityaGarg8:
I am adding this back cause I am now able to get some time to maintain this. Also, it has the 5.10 kernels which are not there in the other marcosfad’s repo.

### DIFF
--- a/docs/guides/wifi.md
+++ b/docs/guides/wifi.md
@@ -66,7 +66,7 @@ This older firmware is archived at [https://packages.aunali1.com/apple/wifi-fw/1
 
 The default kernels for **mbp16-fedora** and **mbp-nixos** use Big Sur firmware.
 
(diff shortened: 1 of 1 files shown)


Redecorating (changes requested):
Can you mention why people would need the kernel from your repo (i.e. for bluetooth on 16,1)?

networkException:
Also alternative link sounds like it's a mirror rather than a separately maintained fork

AdityaGarg8:
Apart from Bluetooth what should I add?

AdityaGarg8:
I guess you may add updated faster, but practically depends on my schedule

AdityaGarg8:
As in I have got a whole week free

AdityaGarg8:
Should I faster?

Redecorating (approved):

PR #177, opened 2021-11-02
Keyboard backlight works on 16.x models

AdityaGarg8:


Redecorating:
It's probably best not to remove that until the fix gets merged into apple-ib-drv, but in the meantime, probably say something along the lines of 

> Keyboard backlight on models MacBookPro16,X and MacBookAir9,1 require a [patched version](link here) of apple-ib-tb which still needs a bit of work before it can be merged into t2linux/apple-ib-drv

That bit of work being putting the backlight stuff into the tb_dev struct, freeing it when the module is unloaded, and ideally not hard coding which models have the backlight through the touchbar backlight device

AdityaGarg8:
Well, we better let this pr remain open then.

AdityaGarg8:
@Redecorating , please review this.

Redecorating (approved):

PR #184, opened 2021-11-28
remove backlash pasted from previous bash command

glutamatt:
https://github.com/t2linux/wiki/commit/b2274b147343aa260d6515f95b25c0b9c4c54300#r61006493

no backslashes before `$` [doc](https://archlinux.org/pacman/pacman.conf.5.html)

### DIFF
--- a/docs/distributions/arch/installation.md
+++ b/docs/distributions/arch/installation.md
@@ -56,7 +56,7 @@ You will need:
 
    ```ini
    [mbp]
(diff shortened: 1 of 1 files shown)


appleplectic:
image

Those may be necessary? They escape the dollar signs, which are needed in the `/etc/pacman.conf` file...

glutamatt:
now the doc has changed and the repo url is not added with bash anymore

image

appleplectic:
Ah you are right

Redecorating:
thanks for catching that!

PR #185, opened 2021-11-29
Mention that fedora now uses Big Sur firmware only

AdityaGarg8:


AdityaGarg8:
@networkException , please review it.

PR #198, opened 2021-12-30
Update the wifi guide for OTP

AdityaGarg8:


docs/guides/wifi.md
```diff
@@ -1,117 +1,32 @@
...
-but make sure to use [https://github.com/jamlam/mbp-16.1-linux-wifi](https://github.com/jamlam/mbp-16.1-linux-wifi)
-instead of [https://github.com/aunali1/linux-mbp-arch](https://github.com/aunali1/linux-mbp-arch)
-as the patchset repository.
-
-### Renaming Firmware
-
-- Rename the `.trx` "Firmware" file to `brcmfmac4364-pcie.bin`, but change `4364` to your wifi chipset number.
-- Do the same for the `.clmb` "Regulatory" file and rename it to `brcmfmac4364-pcie.clm_blob`. Again, change `4364` to your chipset number.
-- Lastly, rename the `.txt` "NVRAM" file to `brcmfmac4364-pcie.Apple Inc.-MacBookPro15,1.txt` but in addition to changing `4364` to your chipset number, change `MacBookPro15,1` to your Model Identifier.
+You may download the firmware from [here](https://github.com/AdityaGarg8/Apple-WiFi-Firmware/releases/download/Montery/firmware.tar.xz).
```
Redecorating, line 7: Can you also include instructions for dumping the firmware from macOS, putting the tarball on the EFI partition, and then extracting it from Linux. That link could get taken down and some people may prefer to get the firmware themselves.

docs/guides/wifi.md
```diff
@@ -1,117 +1,32 @@
...
-1. Now that you got those 3 files, move them to `/lib/firmware/brcm/`.
-2. Check that the files are in place with `ls -l /lib/firmware/brcm|grep -E "43(64|55|77)"`. The output should look something like this
+1. Extract the firmware files from the .tar.xz and move them to `/lib/firmware/brcm/`.
 
-    ```plain
-    -rw-r--r--. 1 root root   12860 Mar  1 12:44 brcmfmac4364-pcie.Apple Inc.-MacBookPro15,1.txt
-    -rw-r--r--. 1 root root  922647 Mar  1 12:44 brcmfmac4364-pcie.bin
-    -rw-r--r--. 1 root root   33226 Mar  1 12:44 brcmfmac4364-pcie.clm_blob
-    ```
+2. Check that the files are in place with `ls -l /lib/firmware/brcm|grep -E "43(64|55|77)"`.
```
Redecorating, line 13: We probably don't need this anymore.

docs/guides/wifi.md
```diff
@@ -1,117 +1,32 @@
...
-- Rename the `.trx` "Firmware" file to `brcmfmac4364-pcie.bin`, but change `4364` to your wifi chipset number.
-- Do the same for the `.clmb` "Regulatory" file and rename it to `brcmfmac4364-pcie.clm_blob`. Again, change `4364` to your chipset number.
-- Lastly, rename the `.txt` "NVRAM" file to `brcmfmac4364-pcie.Apple Inc.-MacBookPro15,1.txt` but in addition to changing `4364` to your chipset number, change `MacBookPro15,1` to your Model Identifier.
+You may download the firmware from [here](https://github.com/AdityaGarg8/Apple-WiFi-Firmware/releases/download/Montery/firmware.tar.xz).
 
 ### Installing Firmware
 
-1. Now that you got those 3 files, move them to `/lib/firmware/brcm/`.
-2. Check that the files are in place with `ls -l /lib/firmware/brcm|grep -E "43(64|55|77)"`. The output should look something like this
+1. Extract the firmware files from the .tar.xz and move them to `/lib/firmware/brcm/`.
```
Redecorating, line 11: Given this will need to be done as root and as some file managers make it hard to elevate to root privileges, could you provide an example of commands needed to extract this to the correct place?

Redecorating (changes requested):
This will also probably have to wait a bit until more T2 distros have kernels with the new WiFi patchset (To speed that up, pull requests to their respective repos would probably help)

docs/guides/wifi.md
AdityaGarg8: Then I'll probably add the way to extract firmware from macOS and remove the link.

AdityaGarg8:
Please review @Redecorating

docs/guides/wifi.md
```diff
@@ -1,120 +1,52 @@
...
-| MacBookAir8,1    | BCM4355 | 12       | Hawaii   | Mojave / Big Sur     |
-| MacBookAir8,2    | BCM4355 | 12       | Hawaii   | Mojave / Big Sur     |
-| MacMini8,1       | BCM4364 | 3        | Lanai    | Mojave / Big Sur     |
-| MacPro7,1        | BCM4364 | 4        | Kahana   | Big Sur              |
-| iMac20,1         | BCM4364 | 4        | Hanauma  | Big Sur              |
-| iMac20,2         | BCM4364 | 4        | Kure     | Big Sur              |
-| iMac19,1*        | BCM4364 | 3        | Midway   | Mojave / Big Sur     |
-| iMac19,2*        | BCM4364 | 3        | Nihau    | Mojave / Big Sur     |
-| iMacPro1,1       | BCM4364 | ?        | Ekans?   | Mojave / Big Sur?    |
+3. Copy the **wifi** folder to an external drive or a partition which is formatted by a file system readable by Linux.
```
networkException, line 13: I'd say "formatted to" and "readable by"

docs/guides/wifi.md
```diff
@@ -1,120 +1,52 @@
...
 
-*If there is missing/uncertain information for your model, please open a [pull request](https://github.com/t2linux/wiki/edit/master/docs/guides/wifi.md) or [issue](https://github.com/t2linux/wiki/issues/new) or mention it on the [discord](https://discord.com/invite/68MRhQu).*
+5. Copy the **wifi** folder to the **Home** folder using your file manager.
 
-### Retrieving Firmware
+6. Open the terminal and run :-
 
-Run `ioreg -l | grep RequestedFiles` in macOS Terminal. Note down the output, and make sure you also know your Model Identifier. The output will be similar to this:
+   ```sh
+   git clone https://github.com/AsahiLinux/asahi-installer --depth=1
```
Redecorating, line 22: I'd prefer to have this bit done on macOS, as that's the process that Marcan had in his [summary patch](https://lore.kernel.org/all/20211226153624.162281-1-marcan@marcan.st/):

> To test this series on a supported Mac today (T2 or M1), boot into macOS
> and run:
>
> $ git clone https://github.com/AsahiLinux/asahi-installer
> $ cd asahi-installer/src
> $ python -m firmware.wifi /usr/share/firmware/wifi firmware.tar
>
> Then copy firmware.tar to Linux and extract it into /lib/firmware.

I think once asahi-installer gets a graphical interface, it'll be doing it from macOS automatically as marcan described, so that could break more of the procedure that we currently have.

docs/guides/wifi.md
```diff
@@ -1,120 +1,52 @@
...
+   cd asahi-installer/src
+   python3 -m firmware.wifi ~/wifi firmware.tar
+   cd /lib/firmware
```
Redecorating, line 25: Then this bit would be where you change to linux

docs/guides/wifi.md
```diff
@@ -1,120 +1,52 @@
...
-| MacBookAir8,1    | BCM4355 | 12       | Hawaii   | Mojave / Big Sur     |
-| MacBookAir8,2    | BCM4355 | 12       | Hawaii   | Mojave / Big Sur     |
-| MacMini8,1       | BCM4364 | 3        | Lanai    | Mojave / Big Sur     |
-| MacPro7,1        | BCM4364 | 4        | Kahana   | Big Sur              |
-| iMac20,1         | BCM4364 | 4        | Hanauma  | Big Sur              |
-| iMac20,2         | BCM4364 | 4        | Kure     | Big Sur              |
-| iMac19,1*        | BCM4364 | 3        | Midway   | Mojave / Big Sur     |
-| iMac19,2*        | BCM4364 | 3        | Nihau    | Mojave / Big Sur     |
-| iMacPro1,1       | BCM4364 | ?        | Ekans?   | Mojave / Big Sur?    |
+3. Copy the **wifi** folder to an external drive or a partition which is formatted to a file system readable by Linux.
```
Redecorating, line 13: I don't think many people will realise they can use the EFI partition for this: `sudo diskutil mount disk0s1`, move firmware to the "EFI" disk

In future, the asahi installer will end up putting the firmware on the efi partition, and distros will eventually expect to be able to fetch it from there. Ideally we'd have the firmware always at something like /firmware.tar on the efi partition, so installer isos and so forth can automatically fetch it and get wifi working, without bundling the firmware in the iso/installers

docs/guides/wifi.md
AdityaGarg8: Consider the fact that not all users use Xcode and Homebrew

docs/guides/wifi.md
AdityaGarg8: Well, for now the guide is written to get wifi working for T2 Macs as per the current situation. We have to option to update it when Asahi releases something, which probably is not coming too soon.

docs/guides/wifi.md
AdityaGarg8: Also, the sole purpose of writing this guide is to get things working as per the current situation. We shall keep it more updated when the situation gets better from Asahi’s side. Probably, the Asahi installer won’t demand git and python for Mac, or they’ll put some documentation to use it which I am not aware about.

Redecorating: I forgot that git isn't installed by default (and i supposed python too). That makes sense then.

docs/guides/wifi.md
Redecorating: Could you still include instructions to use the efi partition for this, as that's probably the most convenient place to use to move the firmware?

AdityaGarg8: But this shall make the things compicated for those who have a triple boot system and use a seperate EFI partition. As I said, this is a guide for the present situation, and not for when Asahi releases the installer.

docs/guides/wifi.md
```diff
@@ -1,120 +1,52 @@
...
-| iMac19,1*        | BCM4364 | 3        | Midway   | Mojave / Big Sur     |
-| iMac19,2*        | BCM4364 | 3        | Nihau    | Mojave / Big Sur     |
-| iMacPro1,1       | BCM4364 | ?        | Ekans?   | Mojave / Big Sur?    |
+3. Copy the **wifi** folder to an external drive or a partition which is formatted to a file system readable by Linux. You can also use your EFI partition by mounting it using `sudo diskutil mount disk0s1`.
 
-\*iMac19,x don't have the T2 chip, but they have the same type of wifi chip as T2 Macs.
+4. Boot into Linux.
 
-*If there is missing/uncertain information for your model, please open a [pull request](https://github.com/t2linux/wiki/edit/master/docs/guides/wifi.md) or [issue](https://github.com/t2linux/wiki/issues/new) or mention it on the [discord](https://discord.com/invite/68MRhQu).*
+5. Copy the **wifi** folder to the **Home** folder using your file manager.
```
Redecorating, line 17: Could you also mention that if they put the firmware on the eif partition, they can access it with `sudo mount /dev/nvme0n1p1 /mnt`

AdityaGarg8: Won't they have to unmount it first from `/boot/efi`(if using it as EFI on Linux too)

Redecorating: Good point, they'd have to check /boot/efi, if its not there then they'd have to mount it

AdityaGarg8: Exactly. This shall make the guide complicated. We shall have to make 2 cases now. A better option seems to be the one I mentioned in the guide. Also, for advanced users, the line about EFI I mentioned seems appropriate enough for them to get it.

Redecorating (approved):

networkException (approved):

AdityaGarg8:
@Redecorating , please review this too.

docs/guides/wifi.md
```diff
@@ -1,120 +1,60 @@
...
+6. Open the terminal and run :-
 
-You will next need to retrieve three of the four files listed by that command (the `.trx` "Firmware" file, the `.clmb` "Regulatory" file, and the `.txt` "NVRAM" file). The source of these files depends on which firmware type you will use. Copy them to somewhere that you can access them.
+   ```sh
+   unzip asahi-installer-main.zip
+   cd asahi-installer-main/src
+   python3 -m firmware.wifi ~/wifi firmware.tar
+   cd /lib/firmware
+   sudo tar xvf ~/asahi-installer-main/src/firmware.tar
+   sudo rm -r ~/asahi-installer-main* ~/wifi
```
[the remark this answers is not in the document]
AdityaGarg8: Not on Ubuntu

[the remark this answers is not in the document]
AdityaGarg8: cd /root required sudo

docs/guides/wifi.md
```diff
@@ -1,120 +1,60 @@
...
-| MacMini8,1       | BCM4364 | 3        | Lanai    | Mojave / Big Sur     |
-| MacPro7,1        | BCM4364 | 4        | Kahana   | Big Sur              |
-| iMac20,1         | BCM4364 | 4        | Hanauma  | Big Sur              |
-| iMac20,2         | BCM4364 | 4        | Kure     | Big Sur              |
-| iMac19,1*        | BCM4364 | 3        | Midway   | Mojave / Big Sur     |
-| iMac19,2*        | BCM4364 | 3        | Nihau    | Mojave / Big Sur     |
-| iMacPro1,1       | BCM4364 | ?        | Ekans?   | Mojave / Big Sur?    |
+   ```sh
+   cp -r /usr/share/firmware/wifi ~/Desktop
+   curl -L https://github.com/AsahiLinux/asahi-installer/archive/refs/heads/main.zip > ~/Desktop/asahi-installer-main.zip
```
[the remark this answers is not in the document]
AdityaGarg8: review now

docs/guides/wifi.md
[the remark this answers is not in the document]
Redecorating: > Not on Ubuntu

If sudo isn't needed in this line, then it shouldn't be there. Both files will be owned by the user, so root won't be required to remove them.

> cd /root required sudo

That would need sudo, but I'm not sure when `cd /root` is being used.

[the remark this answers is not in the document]
AdityaGarg8: > If sudo isn't needed in this line, then it shouldn't be there. Both files will be owned by the user, so root won't be required to remove them

Which files are you taking about?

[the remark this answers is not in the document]
Redecorating: these ones `~/asahi-installer-main* ~/wifi`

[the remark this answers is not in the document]
AdityaGarg8: Without sudo, they seem to give an error stating that they are read-only on Ubuntu.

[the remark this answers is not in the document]
AdityaGarg8: I guess this was some security measure to prevent deletion in the home folder introduced by canonical (not sure though). Anyways, no harm is using sudo though.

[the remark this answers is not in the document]
Redecorating: I think I misinterpreted what you meant by this

> Not on Ubuntu

but if that's the case then including sudo should be fine

networkException (approved):

PR #212, opened 2022-01-16
Fixed formatting of Gentoo guide

EmreTech:
Didn't think this would happen I'm sorry 😖

### DIFF
--- a/docs/distributions/gentoo/installation.md
+++ b/docs/distributions/gentoo/installation.md
@@ -14,18 +14,26 @@
 
     1. Open the Bootcamp installer and follow along until it requests for you to input a Windows ISO. This should clear space for a Linux partition because of removed APFS snapshots.
     2. Now open up Disk Utility. Make a partition with any format. The amount of space you allocate for this Linux partition will be final, so choose wisely.
+
 2. Download a Gentoo ISO and flash it to your USB Flash Drive via [Balena Etcher](https://www.balena.io/etcher/) (you can also use dd)
 3. Ensure that Secure Boot is disabled
+
     1. Follow the instructions from [this article](https://support.apple.com/en-us/HT208198)
     2. Once in the Startup Security Utility, set secure boot to no security and enable external boot.
+
 4. Boot into the Live USB enviornment
+
     1. Plug in the USB Flash Drive into your computer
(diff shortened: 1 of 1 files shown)


AdityaGarg8:
Though it's merged now, I still have some doubts.

AdityaGarg8:
Firstly, we can use disk utility to partition the SSD. Why use Bootcamp?

AdityaGarg8:
Infact bootcamp starts partitioning after getting path of the Windows ISO iirc.

AdityaGarg8:
Secondly, talking about the patches, the additional patch mentioned for kernels below 5.16 is the EFI one which you have included in the mainline patches (Redecorating's fork), then why specially mentioned that one?

AdityaGarg8:
Lastly, the source of kernel being compiled is lastest from kernel.org, or the ones supplied by gentoo?

Redecorating:
> Firstly, we can use disk utility to partition the SSD. Why use Bootcamp?

The introduced guide uses bootcamp assistant to clear snapshots and make space on the macOS apfs partition, so it can then be partitioned with Disk Utility. Bootcamp Assistant itself isn't doing the partitioning.

> Secondly, talking about the patches, the additional patch mentioned for kernels below 5.16 is the EFI one which you have included in the mainline patches (Redecorating's fork), then why specially mentioned that one?

Gentoo is still on 5.15, and if they checkout a 5.15 tag on my fork then it'll not have the efi patch. Adding in that patch means they don't have to mess around with efi=noruntime or mounting efivarfs ro

AdityaGarg8:
Right
Thanks

EmreTech:
> Secondly, talking about the patches, the additional patch mentioned for kernels below 5.16 is the EFI one which you have included in the mainline patches (Redecorating's fork), then why specially mentioned that one?

It's a replacement for making efivars unwritable (suggested by Redecorating themselves)

EmreTech:
> Lastly, the source of kernel being compiled is latest from kernel.org, or the ones supplied by gentoo?

It's provided from Gentoo, the package name is "gentoo-sources". It doesn't have the latest kernel yet, but it has some patches to the kernel that improve security and stuff.

PR #213, opened 2022-01-16
Add new ISO with bugs fixed

AdityaGarg8:
The reason for failures on the previous ISO probably was the old kernel config (way back from 5.6). I've recompiled 5.15.14 using official kernel config for 5.15 from Ubuntu Jammy Jellyfish fork and now users seem to be getting success


Redecorating:
Does the link need to be to a specific release? https://github.com/AdityaGarg8/mbp-ubuntu/releases/latest will redirect to the latest release that isn't a pre-release (if new releases need to be tested then they could be marked as pre-releases)

Redecorating (approved):

PR #218, opened 2022-01-19
Remove obsolete wifi guide step in Arch guide

AdityaGarg8:


EmreTech:
I think it should be kept because since you're still in macOS, it saves time
But I guess you'll only waste like 2-3 minutes, so yeah

Redecorating:
Can we replace it with a suggestion to look at the section in the new wifi guide that is meant to be done on macos?

AdityaGarg8:
Isn't the Arch ISO preloaded with firmware?

Redecorating:
hmm there aren't any instructions on how to copy over firmware files from an iso to a new installation. (unless there's a package file on the iso filesystem for that)

AdityaGarg8:
I've added the link to the wifi guide at the end.

docs/distributions/arch/installation.md
```diff
@@ -83,6 +82,7 @@ You will need:
...
-12. You can follow the [wifi guide](https://wiki.t2linux.org/guides/wifi/) (if you have already retrieved the correct firmware files, you only need to follow the rest of it) now, or after rebooting into your install.
-13. You now will be able to select your Arch install in the macOS Startup Manager by holding option at boot.
+10. Make nvram/efivars automatically remount as readonly, as writing to them causes a panic (deleting and reading variables, however, does not): `echo efivarfs /sys/firmware/efi/efivars efivarfs ro,remount,nofail 0 0 >> /etc/fstab`. If this doesn't work, you can instead add the `efi=noruntime` kernel parameter as described when installing your bootloader (but don't use both of these fixes at the same time).
+11. You can follow the [wifi guide](https://wiki.t2linux.org/guides/wifi/) (if you have already retrieved the correct firmware files, you only need to follow the rest of it) now, or after rebooting into your install.
```
Redecorating, line 86: I forgot that this line is here, but you can probably edit it to be more applicable to the current wifi guide.

Redecorating (changes requested):

docs/distributions/arch/installation.md
```diff
@@ -83,6 +82,7 @@ You will need:
...
+12. You now will be able to select your Arch install in the macOS Startup Manager by holding option at boot.
+13. If Wi-Fi isn't working in the new installation, try following the [Wi-Fi guide](https://wiki.t2linux.org/guides/wifi/).
```
Redecorating, line 88: I think I accidentally didn't save a comment here, but this line is redundant given the one above (although the one above ought to be reworded to fit with the new guide better)

AdityaGarg8: Checkout now

Redecorating (approved):

PR #219, opened 2022-01-20
Update Kernel Compile Guide

Redecorating:
I've tested this process on Arch and Ubuntu (Ubuntu happens to make the initramfs, update grub config and set the new kernel as default when you do `sudo make install`).

### DIFF
--- a/docs/guides/kernel.md
+++ b/docs/guides/kernel.md
@@ -40,8 +40,16 @@ done
 
 ### Setting kernel configuration
 
+There are a few places you could get your kernel config.
+
+- Use the config of the kernel that is currently running (if your running kernel is an older longterm kernel, there could be issues caused by updating its kernel config with the defaults for new options): `zcat /proc/config.gz > .config`
+- Use the Arch Linux kernel config (may have differences to the normal config of your distro, but is up to date): `cp ../patches/config .config`
+
+You can always use `make menuconfig` to change kernel config options later if you have issues.
+
+Then run:
+
 ```bash
-zcat /proc/config.gz > .config
 make olddefconfig
 scripts/config --module apple-ibridge
 scripts/config --module apple-bce
@@ -52,11 +60,10 @@ scripts/config --module apple-bce
 This may take 2-3 hours to build depending on your CPU and the kernel config.
 
 !!! Info "Incremental builds"
-    If you `control-c` to stop the build process, you may continue where you left off by running it again. If you build the kernel, and realise you want to make more changes, re-running `make` will only rebuild bits that you changed.
+    If you `control-c` to stop the build process, you may continue where you left off by running `make` again. If you build the kernel, and realise you want to make more changes to the code or config, re-running `make` will only rebuild bits that you changed.
 
 ```bash
-export MAKEFLAGS=-j$(nproc)
-make
+make -j$(nproc)
 ```
 
 ## Installing
--- a/docs/guides/kernel.md
+++ b/docs/guides/kernel.md
@@ -18,10 +18,14 @@ If your distro is not one of the distros with documentation on this Wiki, you ma
 
 ### Getting kernel source and applying patches
 
+!!! Hint
+    The kernel source will be downloaded with HTTP**S**, but if you would like to be extra careful and verify the kernel source code with `gpg`, please refer to [this page](https://kernel.org/signature.html#using-gnupg-to-verify-kernel-signature).
+
 ```bash
 mkdir build && cd build
 git clone --depth=1 https://github.com/Redecorating/mbp-16.1-linux-wifi patches
 source patches/PKGBUILD
+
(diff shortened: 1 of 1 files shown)


networkException (approved):
LGTM, maybe you could also add a step to check the signature of the kernel source tarball (if thats not too complicated)

docs/guides/kernel.md
```diff
@@ -1,82 +1,110 @@
...
+done
 ```
 
-If you want to install to your current system directly (the kernel will not be managed by your package manager):
+### Setting kernel configuration
 
-```sh
-pkgdir=/
+```bash
+zcat /proc/config.gz > .config
```
AdityaGarg8, line 40: Ubuntu is not a rolling release distro. So, it still uses kernel config from way back 5.4. Thus for Ubuntu you may get the config by :-
```bash
curl -L https://github.com/AdityaGarg8/T2-Ubuntu-Kernel/raw/Mainline/templates/default-config > .config
sed -i 's/CONFIG_VERSION_SIGNATURE=.*/CONFIG_VERSION_SIGNATURE=""/g' .config
sed -i 's/CONFIG_SYSTEM_TRUSTED_KEYS=.*/CONFIG_SYSTEM_TRUSTED_KEYS=""/g' .config
sed -i 's/CONFIG_SYSTEM_REVOCATION_KEYS=.*/CONFIG_SYSTEM_REVOCATION_KEYS=""/g' .config
sed -i 's/CONFIG_DEBUG_INFO=y/# CONFIG_DEBUG_INFO is not set/g' .config
make olddefconfig
scripts/config --module apple-ibridge
scripts/config --module apple-bce
```

AdityaGarg8 (changes requested):

docs/guides/kernel.md
Redecorating: it works fine when I grab the config from your kernel, but i dunno about older kernels and other distros

many distros have different configs so I'm not sure if it would work to have a long list of links to configs for recent kernels for a bunch of distros

If we use the arch config then that won't include, for example, the modules needed for anbox, which might confuse people on distros that normally support it (e.g. ubuntu). Maybe something like this could merge them, depending on how .config is parsed (i'll have to look into that when i get time)

```sh
cp ../patches/config .config
zcat /proc/config.gz >> .config
```

Does it fail to build if debian/canonical-certs.pem isn't there unless you remove CONFIG_SYSTEM_TRUSTED_KEYS? Instead of sed we could probably use scripts/config with --set-str and -u

AdityaGarg8: > it works fine when I grab the config from your kernel, but i dunno about older kernels and other distros
My kernel obviously will have the modified config so you won't face issues. But, some people prefer compiling on PCs having Linux support natively. They shall face issues then.
> 
> many distros have different configs so I'm not sure if it would work to have a long list of links to configs for recent kernels for a bunch of distros
> 
> If we use the arch config then that won't include, for example, the modules needed for anbox, which might confuse people on distros that normally support it (e.g. ubuntu). Maybe something like this could merge them, depending on how .config is parsed (i'll have to look into that when i get time)
> 
> ```shell
> cp ../patches/config .config
> zcat /proc/config.gz >> .config
> ```
> 
> Does it fail to build if debian/canonical-certs.pem isn't there unless you remove CONFIG_SYSTEM_TRUSTED_KEYS? Instead of sed we could probably use scripts/config with --set-str and -u
You may test it out if you get time.

Redecorating: Do we have examples of using older config and then make olddefconfig breaking stuff?

AdityaGarg8: I don't have one. But old configs do cause non compilation of drivers that were upstreamed later, causing performance issues or lack of features. Eg:- I was unable to use exfat for a long time due to old kernel config.

docs/guides/kernel.md
```diff
@@ -40,8 +40,16 @@ done
 
 ### Setting kernel configuration
 
+There are a few places you could get your kernel config.
+
+- Use the config of the kernel that is currently running (if your running kernel is an older longterm kernel, there could be issues caused by updating its kernel config with the defaults for new options): `zcat /proc/config.gz > .config`
+- Use the Arch Linux kernel config (may have differences to the normal config of your distro, but is up to date): `cp ../patches/config .config`
```
AdityaGarg8, line 46: Arch Linux kernel config shows issues with Ubuntu. I tried to compile using it once and ended up having just a terminal at boot.

AdityaGarg8 (changes requested):

docs/guides/kernel.md
Redecorating: I'm not sure how we'd get up to date configs for every distro then, as using one distro's config for another can make it not boot at all.

AdityaGarg8: probably stick to `zcat /proc/config.gz > .config` then

AdityaGarg8 (approved):

AdityaGarg8 (approved):

PR #229, opened 2022-01-27
Kernel update instructions

Redecorating:
distros: FAQs: Add instructions for updating kernels to each distro's faq (except gentoo, i'm not sure what the process for gentoo is/will be).

wifi: tell people to update their kernel if they don't have OTP support

I hope i didn't get any links wrong ;-;


networkException (approved):
Very nice

networkException:
Can you fixup the first commit instead of adding a new one? Very minor nitpicking but keeping the git history tidy would be quite nice

PR #230, opened 2022-01-27
Use script for wifi

AdityaGarg8:


Redecorating:
could we put the readme of that gist into the wifi guide? And possibly the script could be put into the wiki repo, although we might need to make a new folder somewhere for things like that.

AdityaGarg8:
>And possibly the script could be put into the wiki repo, although we might need to make a new folder somewhere for things like that.

We can do that later.

docs/guides/wifi.md
```diff
@@ -13,57 +13,66 @@ Refer to the "Updating Kernel" section on your distro's FAQ for instructions if
...
+### On Linux
 
-   This shall make a folder named **wifi** and a zip archive named **asahi-installer-main.tar.gz** on the macOS Desktop.
+You have two options here:
 
-3. Copy the **wifi** folder and the **asahi-installer-main.tar.gz** archive to an external drive or a partition which is formatted to a file system readable by Linux. You can also use your EFI partition by mounting it using `sudo diskutil mount disk0s1`.
+- The first is to either copy this script to Linux via a USB, download it if you have a wired internet connection, or use some other method to get it to Linux. You can then run the script again from Linux and it will finish setting up Wi-Fi.
 
-4. Boot into Linux.
+- The second method is to run the commands which the script asked you to run on macOS. The commands in most cases should be the following, though you have to stick to the commands asked by the script, in case they turn out to be different.
```
Redecorating, line 34: This is quite confusing and could lead readers to think that the entire code block can't be trusted. It'd be better to tell them that "wifi.sh" on the last line could have a different name if they renamed the script on macOS, so they know what could need changing, and why it would need changing.

AdityaGarg8: d1202ac check

Redecorating (approved):

networkException:
I'd be in favor of putting the script and stuff in the repo from the beginning, fragmentation of documentation was one of the reasons for the wiki.

AdityaGarg8:
Which repo?

AdityaGarg8:
dl.t2linux.org ?

networkException:
This git repository

Redecorating:
I was thinking of putting it in the wiki repo so the script and it's documentation in wifi.md can be kept and edited together. I'm not sure if we'd want that to be in the docs folder, maybe we could make a folder called something like "scripts" or "tools"?

networkException:
I'd be for tools, leaves the possibility for non script stuff in the future

AdityaGarg8:
@networkException , please review

docs/guides/wifi.md
```diff
@@ -13,57 +13,66 @@ Refer to the "Updating Kernel" section on your distro's FAQ for instructions if
...
+## Setting up
 
-In order to get wifi running on Linux, we need to get the firmware from macOS and copy it to Linux with appropriate changes. Thanks to Asahi Linux, who has made the process much simpler. You may follow the steps below to get the firmware and copy it to Linux properly.
+We now use a script which can help you set up Wi-Fi. Follow the instructions below to use this script :-
 
-1. Boot into macOS.
+### On macOS
 
-2. Open the terminal and run :-
+1. Click [here](https://github.com/t2linux/wiki/tree/gh-pages/tools/wifi.sh) to download the script.
```
Redecorating, line 22: I'm not sure if the file would actually end up on the gh-pages branch, so I think it'll need to be ...tree/master/tools...

docs/guides/wifi.md
```diff
@@ -13,57 +13,66 @@ Refer to the "Updating Kernel" section on your distro's FAQ for instructions if
...
+## Setting up
 
-In order to get wifi running on Linux, we need to get the firmware from macOS and copy it to Linux with appropriate changes. Thanks to Asahi Linux, who has made the process much simpler. You may follow the steps below to get the firmware and copy it to Linux properly.
+We now use a script which can help you set up Wi-Fi. Follow the instructions below to use this script :-
 
-1. Boot into macOS.
+### On macOS
 
-2. Open the terminal and run :-
+1. Click [here](https://github.com/t2linux/wiki/tree/master/tools/wifi.sh) to download the script.
```
networkException, line 22: Can we use a relative file path instead? That should make it work offline and on different branches

AdityaGarg8: For eg?

AdityaGarg8: check 3894fcc @networkException

networkException (approved):

PR #231, opened 2022-01-27
Fix lint to script

AdityaGarg8:


AdityaGarg8:
@networkException , will this work, or do we also have to edit the mkdocs.yml as well. The wifi guide is currently useless BTW.

networkException:
It works with mkdocs

AdityaGarg8:
Link is working now.

PR #232, opened 2022-01-27
Make some instructions more clear - wifi and minor fixes to the script.

AdityaGarg8:


AdityaGarg8:
cc @Redecorating

docs/guides/wifi.md
```diff
@@ -22,12 +22,12 @@ We now use a script which can help you set up Wi-Fi. Follow the instructions bel
 1. Click [here](../tools/wifi.sh) to download the script.
 2. Boot into macOS.
 3. Run this script there.
-4. The script will ask you to run some commands or the script itself in Linux after it gets executed successfully.
+4. The script will ask you to run some commands, or the script itself, in Linux after it gets executed successfully.
```
Redecorating, line 25: Can you put the "in Linux after it gets executed successfully" clause first, i.e. something like "after the script is executed successfully, it'll ask you to boot to linux and then either run some commands or run the script itself a second time". This will keep that clause separate from the two listed options, increasing cohesion (imo).

Redecorating (changes requested):

docs/guides/wifi.md
AdityaGarg8: fbdf9e0

docs/tools/wifi.sh
```diff
@@ -52,7 +52,8 @@ case "$os" in
 		if [[ ($input != y) && ($input != Y) ]]
 		then
 			echo "Removing the copy from the EFI partition"
-			sudo rm $mountpoint/wifi.tar.gz $mountpoint/asahi-installer-main.tar.gz $mountpoint/"$script_name"
+			sudo rm $mountpoint/wifi.tar.gz $mountpoint/asahi-installer-main.tar.gz
+			sudo rm $mountpoint/"$script_name" || (sudo rm $mountpoint/*.sh || (echo "Warning! Couldn't remove the copy of the script from the EFI partition. Please remove manually." && sleep 5))
```
Redecorating, line 56: don't remove all .sh files, a user might have put other scripts there.

If the user is meant to remove the file manually, then pause the script until they press enter.

I'm not sure why we are adding this code to ensure the script is deleted. In what situation would it not be deleted by the current code? Do the other files need the same level of care to ensure they are deleted?

Redecorating (changes requested):

docs/tools/wifi.sh
AdityaGarg8: For example :- I run wifi.sh on macOS. This will create a script named wifi.sh in EFI. Now, the user chooses to run the script itself again in Linux. Due to some reason, the script got renamed to, say wifi(1).sh. Now, `rm $mountpoint/"$script_name"` shall be `rm /tmp/apple-wifi-efi/wifi(1).sh`, causing error.

AdityaGarg8: 411ea43

Redecorating: Making it always save the copy of the script on the efi partition as "wifi.sh" and then always removing "wifi.sh" from the efi partition should prevent the possibility of the user having to manually remove it then

Redecorating: and then we wouldn't have to worry about the commands to mount the efi partition and run the script from there changing

Redecorating (approved):

PR #235, opened 2022-02-03
Touch Bar and suspend work

AdityaGarg8:


docs/state.md
```diff
@@ -9,15 +9,15 @@ While Linux is usable on all T2 models, some features are limited due to the lac
...
+- Touch Bar
+- Suspend (It works if [this guide](https://wiki.t2linux.org/guides/dkms/#fixing-suspend) is followed.
 
 ## Partially Working
 
 - Bluetooth: Not working on models with the BRCM4377 WiFi chip (MacBookPro15,4, MacBookPro16,3, MacBookAir9,1), and models with the BRCM4355 WiFi chip (MacBookAir8,1, MacBookAir8,2).
 - Trackpad: Though it is technically working, it is far from the experience on macOS. No force touch or palm rejection. Some models have deadzones on the edges of their trackpads.
-- Touchbar: There is support for the so called simple mode, the same that you would see on Bootcamp Windows for example. Either function keys from 1 to 12 or basic media / brightness control are shown. Sometimes it is unable to change between function keys and media / brightness keys.
 - Audio: With proper configuration audio can work, however it is not stable in some older kernels and switching between speakers and when using the microphone. Microphone volume is low in some Macs.
-- Suspend is very slow to resume (20-40 seconds), and the Touchbar sometimes does not work after resume.
```
Redecorating, line 20: Can we keep the "Very slow to resume bit"? It shouldn't be nearly as slow as it is (iirc the thunderbolt/usb-c ports were taking too long according to https://github.com/intel/pm-graph )

docs/state.md
```diff
@@ -9,15 +9,15 @@ While Linux is usable on all T2 models, some features are limited due to the lac
 - USB
 - Keyboard
 - Camera
-- Wifi (For now its working for all the devices we have data about. There are still some devices of whom we don't have data. Refer to the [wifi guide](https://wiki.t2linux.org/guides/wifi/) for more details)
+- Wi-Fi
+- Touch Bar
```
Redecorating, line 13: Maybe still mention that it's the same as the windows driver, no custom graphics.

Redecorating (changes requested):

docs/state.md
AdityaGarg8: 4d7e402

docs/state.md
AdityaGarg8: 4d7e402

Redecorating: Could you also give a number of seconds?

AdityaGarg8: How much?
Takes around 5-10 for me since the touchbar got fixed

Redecorating: I haven't tried suspend since that was fixed, I'll go check if that has improved it

Redecorating: with one test i got 16 seconds

AdityaGarg8: I got 9 rn

Redecorating: looks like it's not as slow then!

> its too slow

maybe "it's very slow (5-20 seconds)". "Too" makes me think this slowness breaks something, but it's just slow.

AdityaGarg8: ed3ce16

Redecorating (approved):

PR #247, opened 2022-02-11
Hybrid graphics: Mention using mount point of rEFInd partition for os set

AdityaGarg8:


Redecorating (commented):
> If you mount the EFI partition somewhere else or use refind, you will need to replace /boot/efi with the mount point of the partition in which your bootloader is installed.

Maybe combine the two sentences with something similar to this?

Redecorating (approved):

PR #251, opened 2022-02-24
state: add detail to some points and some rewording

Redecorating:
Headphone control buttons seem to work consistently for me now that hid
no longer upsets apple-bce by sending debug configuration requests to
the touchbar.


AdityaGarg8 (approved):

AdityaGarg8:
So you are able to monitor usb packets on touchbar on Linux?

Redecorating:
i can use wireshark to capture the usb traffic to and from the touchbar if that's what you mean

the configuration names were from lsusb but I think you can also read them from sysfs

PR #257, opened 2022-03-10
Roadmap: Add warning about removing macOS and possible way to restore.

AdityaGarg8:


networkException (approved):
LGTM

docs/roadmap.md
```diff
@@ -12,8 +12,8 @@ Your first consideration should be the risk you are taking. Don't worry, nobody
 Linux and by following the guides closely you should be able to get everything working even without a lot of knowledge as well.
 Still in case something goes wrong we are not responsible for it, we will try our best to help out however.
 
-Note that while its technically possible to remove macOS entirely, its strongly encouraged to dual boot it with Linux to have
-a backup in case something goes wrong.
+!!! warning "Removing macOS"
+           Note that while its technically possible to remove macOS entirely, its strongly encouraged to dual boot it with Linux to have a backup in case something goes wrong. We shall also recommend [creating a bootable macOS installer](https://support.apple.com/en-us/HT201372) so as to have an option of restoring macOS back.
```
Redecorating, line 16: Keeping macOS is also good for firmware updates.

Also for wiki style ([this](https://en.m.wikipedia.org/wiki/Wikipedia:Manual_of_Style#First-person_pronouns) is better at explaining what I mean), "We shall also recommend" could be passive voice (i.e. "It is also recommended to create a..."). 

You could mention that internet recovery can be slow and unreliable (which is the reason that creating a macOS installer is useful ).

networkException: Removing all the "shall" wording is probably better scoped for a separate pr if that's something we want to pursue

docs/roadmap.md
```diff
@@ -12,10 +12,20 @@ Your first consideration should be the risk you are taking. Don't worry, nobody
...
+While its technically possible to remove macOS entirely, its strongly encouraged to dual boot it with Linux because :-
+
+1. It acts as a backup in case something goes wrong.
+2. macOS updates often bring along certain firmware updates, which tend to be useful for Linux as well.
+
+Still if you wish to remove macOS completely, it is recommended to [create a bootable macOS installer](https://support.apple.com/en-us/HT201372) so as to have an option of restoring macOS back.
+
+You can also use the [Internet Recovery](https://support.apple.com/en-in/HT204904) to reinstall macOS, but this has been quite unreliable and slow for a considerable amount of users, thus making the bootable installer method a better one.
+
+## What all shall work on Linux?
```
Redecorating, line 26: I'm not sure if "all" works here, maybe something like "What will work on Linux" or "What works on Linux" could be better.

docs/roadmap.md
```diff
@@ -12,10 +12,20 @@ Your first consideration should be the risk you are taking. Don't worry, nobody
...
+
+Take a look at the [state article](https://wiki.t2linux.org/state/). It gives a list of roughly what works on Linux and what not.
```
Redecorating, line 28: I think "what not" should be "what doesn't".

Redecorating (approved):
Just some grammar nitpicking, but otherwise LGTM

PR #259, opened 2022-03-12
Typos & old link changed

NoaHimesaka1873:


AdityaGarg8 (commented):
BTW, since we have a dedicated script for the touchbar, is the *Switch touchbar to function keys* section still required?

PR #270, opened 2022-04-04
Make Arch use & install NetworkManager

NoaHimesaka1873:


Redecorating:
What's `iwd` doing? For me at least, both `iwd` and `wpa_supplicant` work fine.

hellosway:
`wpa_supplicant` also works for me. I don't think NetworkManager should be preinstalled, it doesn't fit with the arch philosophy. It is not needed for wifi to work, so it shouldn't be installed, its not in the normal arch installation.

Redecorating:
Maybe we should have a table of  suggested packages for users to add to their pacstrap? That'd have the package name, and why you might need it. This could be something like:

| package | use/reason |
|---------|------------|
|iwd or networkmanager | A wireless networking daemon (needed to connect to wifi)|

Redecorating:
> Ah there's also wpa_supplicant.

wpa_supplicant is the default backend for NetworkManager, so it'll already be in use by default when using NetworkManager. (The NetworkManager backend can be changed to iwd, and wpa_supplicant can be used without NetworkManager.)

>  iwd _rarely_ works

I can't reproduce that on my machine. Testing with [this release](https://github.com/t2linux/archiso-t2/releases/tag/2022.04.02) of archiso-t2 on MBP16,1, connecting with NetworkManager's `nmtui` works fine (wpa_supplicant as the backend). Using that wifi connection to download iwd and then connect to wifi with `iwdctl` also worked fine:

```
nmtui
# connect
ping archlinux.org
# ping worked fine
pacman -Sy iwd
systemctl stop NetworkManager
systemctl stop wpa_supplicant
systemctl start iwd
iwctl
# in iwd shell used `station <wlanX> connect <SSID>`
ping archlinux.org
# ping worked fine with this connection too
```

We're both on the same model so I don't know what would be causing iwd to work fine for me but not for you. Are there logs from iwd not working with the newer wifi patchset?

I neglected to try testing if setting iwd as network manager's backend works on the live iso, maybe that could be the issue? I know it was working for me a few months ago, but I don't know about recently.

hellosway:
> Maybe we should have a table of suggested packages for users to add to their pacstrap? That'd have the package name, and why you might need it. This could be something like:

Since you are supposed to follow the arch wiki, if you read it there is a section that tells you install a network manager, man pager, and some other stuff, that lists the package names.

PR #271, opened 2022-04-06
Fix sleep and resume on Arch Linux

redstrate:
This fixes sleep/resume on Arch Linux, otherwise the keyboard and touchbar fail to wake up properly.

I'm not entirely sure why this works, but without adding the lines for `hid_apple`, the touchbar is permenantly black and the keyboard fails to wake up from sleep. I got this idea from someone on Discord who found this was identical to the script in the Fedora image, so I gave it a try and it worked perfectly (apart from the known issue with the keyboard backlight).

### DIFF
--- a/docs/tools/rmmod_tb.sh
+++ b/docs/tools/rmmod_tb.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 if [ "${1}" = "pre" ]; then
         modprobe -r apple_ib_tb
+        modprobe -r hid_apple
 elif [ "${1}" = "post" ]; then
+        modprobe hid_apple
         modprobe apple_ib_tb
 fi


Redecorating:
Do you have logs (`journalctl -k`) from when the touchbar and keyboard were dead after resume from sleep? There may have been a desync with apple-bce, it's interesting that hid_apple seems to be causing it for you.

redstrate:
Yes I’ll post them as soon as I encounter the bug again :-) Currently it seems like my keyboard has a mind of its own, sometimes it will work perfectly even when resuming - but other times the touchbar will fail to load at all (the modules are loaded however). This happens every bootup, and there doesn’t seem to be a way to force “reload” the touchbar, even with unloading and reloading modules if i encounter a black touchbar.

Im gonna draft this for now though, since it doesn’t completely fix my issue. I’m interested in fixing this because this is really one of the very few pain points when using Linux on this laptop. When I boot back into Arch Linux I’ll amend this with more details :-)

jeffgus:
I'm running Fedora. As mentioned above, Fedora comes with the script to remove and load the modules, BUT it hasn't been working for me. Keyboard/touchpad/touchbar do not work after resume. Ticket shows closed, but I can't find where this has been resolved.

Redecorating:
Do you have logs after it resumes without keyboard etc working? `journalctl -k > current_boot.log` or `journalctl -b -1 > previous_boot.log` can be used to get logs from the current or previous boot.

PR #274, opened 2022-04-16
Added a systemd service to monitor and keep mic volume at 400% on mbp16.1 2019

mahboobkarimian:
If you are using pulseaudio and you can't replace it with pipewire to use its solutions to maintain the volume of the microphone on your Mac, or you use gnome and you can't use the volume increase available in KDE, this script and a systemd service will help you to always maintain your mic volume at 400%. This script is for MacBook Pro 16" 2019, but with a little effort, it can be ported to other Macs who are suffering from low microphone volume or sudden drops during online video/audio calls.
The script continuously checks for mic volume changes using `pactl subscribe` and sets it to 400% when there is a change.


networkException:
LGTM, thank you

PR #278, opened 2022-04-25
Wifi :- Add workaround to fix the no wifi bug on Ubuntu

AdityaGarg8:


networkException:
If there's an issue I haven't spotted we'll just change it again

AdityaGarg8:
> If there's an issue I haven't spotted we'll just change it again

No worries :)

PR #280, opened 2022-04-26
upgrade Ubuntu

mahboobkarimian:
Instruction for upgrading to newer LTS releases, and troubleshooting/recovery in case of failure.

### DIFF
--- a/docs/distributions/ubuntu/upgrade.md
+++ b/docs/distributions/ubuntu/upgrade.md
@@ -1,0 +1,55 @@
+# Upgrade
+
+Recommendations:
+
+* Perform upgrade from LTS to LTS.
+
+Warnings:
+
+* If you have important data and workspace, make sure that you have backup.
+* Make sure you have enough disk space.
+* **Ensure that you met all the below requirements so that you can recover in case of failed upgrade.**
+
+# Hardware Requirements
+
+* USB-C to Ethernet adapter **OR** Smartphone capable of USB tethering to provide internet in case of emergency
+* A cable that can connect one of the above objects to your Mac
+
+# Kernel Requirements
+
+Check the latest official Ubuntu release and find its kernel version. You need to install the same or higher custom kernel version for your current Ubuntu.
+
+# Install Procedure
+
+Most probably you are using a LTS version of the Ubuntu right now. If a newer LTS Ubuntu is released in April, you will not be notified until Canonical 
+release the first point release (e.g. 22.04.1) which normally out around August. However, you can still upgrade to the final version before August.
+1. Open a terminal and:
+
+    1. If it's after first point release, do `sudo do-release-upgrade`.
+
+    2. If it's before and the upgrade tool says there is no newer version, try `sudo do-release-upgrade -d`.
+  
+2. Follow the instruction given by the tool and answer if it asks question about some of the packages that you want to be upgraded, etc.
+3. Let the tool download finish and begin the process. Have an eye in the screen to see how it is going. In the middle it might ask you to chose some options.
+4. If the installation finish successfully, do a reboot and boot your Mac using the current custom kernel you installed before using the instructions available in the wiki. Otherwise, see the next section.
+
+# Troubleshooting
+
+1. In case the upgrade process gives error about broken packets, do not reboot. Try to resolve the problem first, and resume the process.
+2. If your Mac is crashed and you must press and hold power button, then **probably** you will boot to a black screen with minimal console because gnome-shell (default desktop) package is not installed yet. Here you need to deal with all the errors. 
+The goal is to install at least a GUI at first place so that you can get rid of the black screen with tiny text, then try to resolve other problems using GUI.
+
+    1. Login with your user and password.
+
+    2. Connect your Mac to the internet using USB-C to Ethernet adapter or use your smart phone's USB tethering.
+        * `$ ifconfig -a`, it will show all the available network interfaces. remember the one appears when you connect the USB.
+        * bring up the USB interface `$ sudo ifconfig <USB_INTERFACE> up`
+        * Ask for an IP using `$ dhclient`
+        * Ping `$ ping google.com` to make sure you have access to internet
+
+    3. Try `$ sudo apt install --fix-broken`. If it succeeds, resume upgrade and `$ reboot`. If not, go to 4.
+    4. Try `$ sudo apt install gnome-shell` if it succeeds, do `$ reboot`. If not:
+        * Try to read the errors and find out the solution. For example if there is conflict between package version, try to remove the installed one by `$ sudo apt autoremove <PACKAGE_NAME>` and the try install gnome desktop again.
+        * If you really stuck with installing gnome desktop, try other desktops like: `sudo apt install ubuntu-mate-desktop`
+
+    4. Boot to newly installed desktop and do `sudo apt upgrade` and hope it would end up with success, otherwise try to understand the error and resolve them. Always google your errors.
--- a/docs/distributions/ubuntu/upgrade.md
+++ b/docs/distributions/ubuntu/upgrade.md
@@ -44,11 +44,11 @@ The goal is to install at least a GUI at first place so that you can get rid of 
     1. Login with your user and password.
 
     2. Connect your Mac to the internet:
-    
+
         A. Using Wifi:
         * Check if Wifi module is loaded with `$ lsmod | grep brcmfmac`, if not load it with `$ sudo modprobe brcmfmac`.
         * Use Wifi cli tools to connect to your Wifi network. Further instruction could be found on: https://wiki.archlinux.org/title/Network_configuration/Wireless#Authentication
-    
+
         B. Using USB-C to Ethernet adapter or use your smart phone's USB tethering:
         * `$ ifconfig -a`, it will show all the available network interfaces. remember the one appears when you connect the USB.
         * bring up the USB interface `$ sudo ifconfig <USB_INTERFACE> up`
--- a/docs/distributions/ubuntu/upgrade.md
+++ b/docs/distributions/ubuntu/upgrade.md
@@ -19,7 +19,7 @@ For troubleshooting you need internet access. It is possible that Wi-Fi may also
 
 # Kernel Requirements
 
-Before upgrade, check the latest official Ubuntu release and find its kernel version. You need to install the same or higher custom kernel version for your current Ubuntu to make sure that it will work properly without problems related to the kernel version.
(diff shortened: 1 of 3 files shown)


docs/distributions/ubuntu/upgrade.md
```diff
@@ -0,0 +1,55 @@
...
+
+* Perform upgrade from LTS to LTS.
+
+Warnings:
+
+* If you have important data and workspace, make sure that you have backup.
+* Make sure you have enough disk space.
+* **Ensure that you met all the below requirements so that you can recover in case of failed upgrade.**
+
+# Hardware Requirements
```
AdityaGarg8, line 13: Why Ethernet? Doesn't wifi work?

docs/distributions/ubuntu/upgrade.md
```diff
@@ -0,0 +1,55 @@
...
+
+* USB-C to Ethernet adapter **OR** Smartphone capable of USB tethering to provide internet in case of emergency
+* A cable that can connect one of the above objects to your Mac
+
+# Kernel Requirements
+
+Check the latest official Ubuntu release and find its kernel version. You need to install the same or higher custom kernel version for your current Ubuntu.
```
AdityaGarg8, line 20: I don't think we need to install the custom kernel again. It shouldn't be removed imo.

docs/distributions/ubuntu/upgrade.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Install Procedure
+
+Most probably you are using a LTS version of the Ubuntu right now. If a newer LTS Ubuntu is released in April, you will not be notified until Canonical release the first point release (e.g. 22.04.1) which normally out around August. However, you can still upgrade to the final version before August.
+
+1. Open a terminal and:
+
+    1. If it's after first point release, do `sudo do-release-upgrade`.
```
AdityaGarg8, line 28: Probably add the we can use the software updater here as well

docs/distributions/ubuntu/upgrade.md
```diff
@@ -0,0 +1,55 @@
...
+
+    2. If it's before and the upgrade tool says there is no newer version, try `sudo do-release-upgrade -d`.
+  
+2. Follow the instruction given by the tool and answer if it asks question about some of the packages that you want to be upgraded, etc.
```
AdityaGarg8, line 32: Instructions, not instruction

docs/distributions/ubuntu/upgrade.md
```diff
@@ -0,0 +1,55 @@
...
+3. Let the tool download finish and begin the process. Have an eye in the screen to see how it is going. In the middle it might ask you to chose some options.
```
AdityaGarg8, line 33: Tool finish the download

docs/distributions/ubuntu/upgrade.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Troubleshooting
+
+1. In case the upgrade process gives error about broken packets, do not reboot. Try to resolve the problem first, and resume the process.
+2. If your Mac is crashed and you must press and hold power button, then **probably** you will boot to a black screen with minimal console because gnome-shell (default desktop) package is not installed yet. Here you need to deal with all the errors.
+The goal is to install at least a GUI at first place so that you can get rid of the black screen with tiny text, then try to resolve other problems using GUI.
+
+    1. Login with your user and password.
+
+    2. Connect your Mac to the internet using USB-C to Ethernet adapter or use your smart phone's USB tethering.
```
AdityaGarg8, line 44: Again, why not WiFi?

AdityaGarg8 (changes requested):

docs/distributions/ubuntu/upgrade.md
mahboobkarimian: No, `ifconfig -a` didn't show wifi interface. `iwconfig`, `ip link`, etc. I couldn't use `wpa_supplicant` or `iw` to connect.

docs/distributions/ubuntu/upgrade.md
mahboobkarimian: I think we need to have custom_kernel_version >= new_lts_ubuntu_kernel_version to ensure that after the upgrade, the new ubuntu will work without problem. I assume the new ubuntu release doesn't work well with older kernels versions.
Previously installed kernels are preserved.

docs/distributions/ubuntu/upgrade.md
mahboobkarimian: Oh, I forgot to mention that GUI. I will add.

docs/distributions/ubuntu/upgrade.md
mahboobkarimian: But I saw `brcmfmac` was loaded.

AdityaGarg8: What was `uname -r` after upgrading to 22.04

AdityaGarg8: That's kinda strange, cause after installing a custom kernel, technically there is no difference between a T2 Mac and a general PC, provided we load the kernel.

Is it possible that in your case the kernel was quite old?

mahboobkarimian: `uname -r` shows `5.14.16-t2-big-sur` after upgrade.
Seems Ubuntu 22.04 is shipped with 5.15 and mine was a little old, this is why I recommend users to upgrade their kernel to >= shipped version.
We can add using wifi in this doc, but I am not sure where we need to write it, and the exact instructions. Could someone else contribute to this?

mahboobkarimian: If you check the Kernel Requirements section, I've mentioned it.

AdityaGarg8: Then I don't think we need to specify Ethernet. In your case, the kernel booting wasn't correct.

mahboobkarimian: I put it in the Troubleshooting section, as additional info for users who might face similar problems (for any reason).

docs/distributions/ubuntu/upgrade.md
mahboobkarimian: I added a few words to the wiki to make this more clear.

docs/distributions/ubuntu/upgrade.md
mahboobkarimian: Added "using Wifi".

docs/distributions/ubuntu/upgrade.md
```diff
@@ -41,7 +43,13 @@ The goal is to install at least a GUI at first place so that you can get rid of
 
     1. Login with your user and password.
 
-    2. Connect your Mac to the internet using USB-C to Ethernet adapter or use your smart phone's USB tethering.
+    2. Connect your Mac to the internet:
+    
+        A. Using Wifi:
+        * Check if Wifi module is loaded with `$ lsmod | grep brcmfmac`, if not load it with `$ sudo modprobe brcmfmac`.
+        * Use Wifi cli tools to connect to your Wifi network. Further instruction could be found on: https://wiki.archlinux.org/title/Network_configuration/Wireless#Authentication
```
AdityaGarg8, line 50: Why not GUI?

AdityaGarg8 (changes requested):

Redecorating:
If you are adding a new page, don't forget to add it to `mkdocs.yml` in the root folder of the repo.

docs/distributions/ubuntu/upgrade.md
AdityaGarg8: My bad

docs/distributions/ubuntu/upgrade.md
```diff
@@ -0,0 +1,63 @@
...
+
+Warnings:
+
+* If you have important data and workspace, make sure that you have backup.
+* Make sure you have enough disk space.
+* **Ensure that you met all the below requirements so that you can recover in case of failed upgrade.**
+
+# Hardware Requirements
+
+In general, for troubleshooting you may need Internet access. If Wifi would work after a failure in upgrade process you **DO NOT** need the following requirements.
```
AdityaGarg8, line 15: This part though correct,now very confusing to read. You could firstly move this to the troubleshooting section instead of making this a seperate one. Secondly, you could make it simple by write something like :-

For troubleshooting you need internet access. It is possible that Wi-Fi may also not work for some users during troubleshooting. In that case, you shall also require <requirements of wired net>.

AdityaGarg8 (changes requested):

docs/distributions/ubuntu/upgrade.md
```diff
@@ -0,0 +1,63 @@
...
+1. In case the upgrade process gives error about broken packets, do not reboot. Try to resolve the problem first, and resume the process.
+2. If your Mac is crashed and you must press and hold power button, then **probably** you will boot to a black screen with minimal console because gnome-shell (default desktop) package is not installed yet. Here you need to deal with all the errors.
+The goal is to install at least a GUI at first place so that you can get rid of the black screen with tiny text, then try to resolve other problems using GUI.
+
+    1. Login with your user and password.
+
+    2. Connect your Mac to the internet:
+
+        A. Using Wifi:
+        * Check if Wifi module is loaded with `$ lsmod | grep brcmfmac`, if not load it with `$ sudo modprobe brcmfmac`.
```
AdityaGarg8, line 49: You have used a `$` in front of every command here. I don't think that's required, since it's already there in the terminal.

docs/distributions/ubuntu/upgrade.md
```diff
@@ -0,0 +1,63 @@
...
+        * Use Wifi cli tools to connect to your Wifi network. Further instruction could be found on: https://wiki.archlinux.org/title/Network_configuration/Wireless#Authentication
```
AdityaGarg8, line 50: Also, you could replace Wifi with Wi-Fi. (Probably use `sed` for the same)

AdityaGarg8 (changes requested):
Apart from these, add this to the `mkdocs.yml`.

AdityaGarg8 (approved):

AdityaGarg8:
Though it looks good now, but could we an alternate way of getting 22.04, i.e. upgrade to 21.04 first and then to 22.04?

I actually had followed this way

mahboobkarimian:
> Though it looks good now, but could we add an alternate way of getting 22.04, i.e. upgrade to 21.04 first and then to 22.04?
> 
> I actually had followed this way

Well, that's another way. Canonical recommends LTS to LTS upgrade, at least for servers.
I didn't find any documentation that they say about upgrading from LTS to non-LTS. So personally, I follow what they recommend.
But I think you can do it. I don't know the possible complications, and of course, you do it twice which consumes double time/energy.

PR #306, opened 2022-08-09
Contribute: Add GitHub Sponsors link for networkException

networkException:


AdityaGarg8:
Could you add @mikeeq as well

networkException:
I'd rather they'd add it themselves

PR #308, opened 2022-08-14
Update Fan guide 

thepolarbear0818:
I updated the fan guide to tell users to copy the config file to /etc, since you need to have the config there, and if you dont, sudo make tests will fail 

@networkException  Can you take a look at this and make sure that It is correct? Thanks


networkException:
I think it would be better to teach the test runner to use the file from git rather than having to install it first. What do you think? See https://github.com/networkException/mbpfan/issues/6

thepolarbear0818:
I flipped the order to copy first, then edit later if needed
@networkException

networkException:
> I could be completely wrong

you are, I want have `sudo make tests` work without any copying to `/etc`. It doesn't really make sense to edit the config for some tests.

thepolarbear0818:
Ok, the better solution would be to add it into your repo, but for now I am going to keep this pull request open until it is added.

PR #318, opened 2022-08-30
Changed command suggestion from `dd` to `cp`

obsidianical:
`cp` is generally more intuitive and better suited for flashing disks then `dd`.
Additionally, the [official NixOS installation guide](https://nixos.wiki/wiki/NixOS_Installation_Guide) seems to recommend `cp` over `dd`. So does the arch wiki.

### DIFF
--- a/docs/distributions/nixos/installation.md
+++ b/docs/distributions/nixos/installation.md
@@ -6,7 +6,7 @@
 
 ## Installation Steps
 
-1. Get a NixOS iso from the releases page of [this repo](https://github.com/kekrby/nixos-t2-iso) and write it to your USB using `dd` or another tool.
+1. Get a NixOS iso from the releases page of [this repo](https://github.com/kekrby/nixos-t2-iso) and write it to your USB using `cp` or another tool.
 2. Partition your disk using Disk Utility by splitting the main APFS partition into two partitions where one of them is a FAT32 partition with the amount of space you want to allocate for your Linux installation.
 3. [Open the Startup Security Utility](https://support.apple.com/en-us/HT208198) and set the security level to "no security".
 4. Reboot while pressing the `Option` key and select the orange `EFI Boot` option.


Redecorating:
It looks like for macos the nixos install guide [suggests dd](https://nixos.wiki/wiki/NixOS_Installation_Guide#From_macOS) (but for linux they indeed suggest cp).

Arch is the same, [suggesting dd for macos](https://wiki.archlinux.org/title/USB_flash_installation_medium#In_macOS), but for linux it seems like there's [5 tools they suggest](https://wiki.archlinux.org/title/USB_flash_installation_medium#Using_basic_command_line_utilities), with cat being the first

obsidianical:
`cp` seems the most intuitive and clear to me in that case, and that part might be outdated. `dd`, `cat`, `cp`, etc. all can do the same anyway and are used for the same, and `dd` has no advantage over `cp` while having the risk of the i and the o on the keyboard being close together, making it very possible to accidentally wipe your disk just because of a typo.

Redecorating:
Have you checked that the `cp`included with macOS works correctly when copying to a /dev/diskX file? I do find it strange that they both have the same different advice for macOS compared to Linux so there might be a reason.

> that part might be outdated

In that case I'd suggest updating the nixos wiki and arch wiki pages, and then once that happens we can change the t2wiki without giving advice different to upstream (but I would suggest checking that cp does work on macos for this if you haven't already)

obsidianical:
I'm not able to check if `cp` works the same way on macos as I don't have macos installed anymore, but I don't see why apple would've changed how `cp` behaves.

Redecorating:
You might be able to check from internet recovery? My concern with cp is that it might interact with block special files like /dev/disk0 differently under a Darwin kernel compared to a Linux kernel.

Regardless I think the t2 wiki should stick to the advice the upstream distros have where possible (which means that if upstream distros change their mind, then we should change the t2 wiki to reflect that)

obsidianical:
Then someone (I'll try later if noone did it until then) may check it. I think that the NixOS installation guide makes it pretty visible that `cp` is preferred.

kekrby:
I just tried this and while `cp` can successfully write an iso (I did not try to boot though), it is **very** slow compared to `dd`. I think `dd` should stay recommended.

obsidianical:
Okay, then I agree, thanks for testing it out!

PR #320, opened 2022-09-04
Added required second fan notes to fan.md

lemmyg:
added additional info required to control the second fan.


thepolarbear0818:
I think you're missing an "a" in "Note for laptops with second fan"

lemmyg:
Note that for laptops with a second fan, you need to add `min_fan2_speed` and `max_fan2_speed`?

lemmyg:
> I think you're missing an "a" in "Note for laptops with second fan"

done

PR #325, opened 2022-09-14
docs/distributions/nixos: add instructions on how to set up firmware

kekrby:


docs/distributions/nixos/installation.md
```diff
@@ -7,16 +7,46 @@
...
+    sudo modprobe -r brcmfmac && sudo modprobe brcmfmac
+    ```
+
+    Then run `systemctl start wpa_supplicant` and then connect to internet using `wpa_cli`.
+    When running the commands, don't forget to change `/mnt/boot` to the place you mounted your ESP at (for example `/mnt/boot/efi`).
+
+8. Generate your configuration using `sudo nixos-generate-config --root /mnt`.
+9. Edit `/mnt/etc/nixos/configuration.nix`:
+    * Add `"${builtins.fetchGit { url = "https://github.com/kekrby/nixos-hardware.git"; }}/apple/t2"` to `imports`.
+    * Copy the WiFi/Bluetooth firmware to `/mnt/etc/nixos/firmware/brcm` and add this snippet to your `configuration.nix`:
```
AdityaGarg8, line 32: Doesn't runing the script itself in nixOS copy the firmware correctly?

kekrby: Firmware is installed in NixOS a bit differently, there is no `/lib/firmware` (you can create it, and it is actually done in this tutorial to get WiFi working in the installer, but it is not really how things are done normally). This is the "correct way" of loading firmware.

AdityaGarg8: Ok then

mikroskeem: I think this is one of the cleanest approaches you can do. Something similar woould work with flakes as well.

mikroskeem (approved):

PR #328, opened 2022-09-18
Audio Config: Add tips for the platforms with GRUB2 installed.

imichael2e2:
Some platforms have shipped with GRUB 2.x instead of GRUB 1.x(e.g., Fedora 36), and the command line interfaces for 2.x are slightly different from that of 1.x thus it might be appropriate to add more descriptions here.

Tested on: Fedora36, MBP2019.

Anyway, really appreciate that you guys make my laptop usable, thanks!

### DIFF
--- a/docs/guides/audio-config.md
+++ b/docs/guides/audio-config.md
@@ -12,6 +12,7 @@ If not present, you'll have to update your bootup kernel params:
 
 - edit `/etc/default/grub` and update `GRUB_CMDLINE_LINUX` to include `intel_iommu=on iommu=pt pcie_ports=compat`
 - Apply your edits by running `sudo update-grub` on ubuntu or `sudo grub-mkconfig -o /boot/grub/grub.cfg` for other distros
+  - *Note* that, on the platforms where `grub`'s version is 2.x, these commands should be instead `sudo update-grub2` or `sudo grub2-mkconfig -o /boot/grub2/grub.cfg`. Check `grub`'s version with `sudo grub-install --version` or `sudo grub2-install --version`
 - Reboot and ensure `cat /proc/cmdline` contains those params
 
 !!!note "systemd-boot"


docs/guides/audio-config.md
```diff
@@ -12,6 +12,7 @@ If not present, you'll have to update your bootup kernel params:
 
 - edit `/etc/default/grub` and update `GRUB_CMDLINE_LINUX` to include `intel_iommu=on iommu=pt pcie_ports=compat`
 - Apply your edits by running `sudo update-grub` on ubuntu or `sudo grub-mkconfig -o /boot/grub/grub.cfg` for other distros
+  - *Note* that, on the platforms where `grub`'s version is 2.x, these commands should be instead `sudo update-grub2` or `sudo grub2-mkconfig -o /boot/grub2/grub.cfg`. Check `grub`'s version with `sudo grub-install --version` or `sudo grub2-install --version`
```
Redecorating, line 15: Not all platforms with grub version two have "2" in the command names (e.g. arch), so it's only some distros that need the alternate command names with "2" added.

Arch Linux:
```
> grub-install --version
grub-install (GRUB) 2:2.06.r322.gd9b4638c5-4
```

Also, I don't know if update-grub2 is a thing anywhere, update-grub command is ubuntu specific, and ubuntu doesn't add "2" to the command names iirc.

Maybe direct users to try grub-xxxx, and if it isn't found then tell them to try grub2-xxxx (unless `grub-install` on fedora is grub version 1)

Redecorating (changes requested):

docs/guides/audio-config.md
imichael2e2: @Redecorating Thanks for reviewing! That was my typo and I just have edited and added more appropriate descriptions and guides.

Redecorating (approved):

PR #329, opened 2022-09-19
Deprecate Manjaro

NoaHimesaka1873:


docs/distributions/manjaro/deprecation.md
```diff
@@ -0,0 +1,15 @@
+# Why is Manjaro deprecated?
+
+Read [This.](https://github.com/arindas/manjarno)
+
+TL;DR: It's poorly maintained and dangerous.
```
Redecorating, line 5: Especially here, I feel like it may be getting a little too opinionated for a wiki ([this](https://en.wikipedia.org/wiki/Wikipedia:Neutral_point_of_view) explains what i mean better than i can).

Maybe saying T2 manjaro will become deprecated and have no maintainer due to these concerns(manjarno link), a good alternative is T2 EndeavourOS would be a bit more neutral?

docs/distributions/manjaro/home.md
```diff
@@ -1,3 +1,7 @@
-Welcome to the mbp-manjaro wiki!
+# Deprecation Notice
+
+Manjaro support is deprecated. See [Deprecation Plan](https://wiki.t2linux.org/distributions/manjaro/deprecation) for more information.
```
Redecorating, line 3: > Manjaro support is deprecated

Not yet from my understanding (this applies to the other pages with this note as well)

AdityaGarg8:
Imo this PR is not required at all. Looking at this, it seems like we are pursuading users to not use Manjaro. Our aim is to provide what the user wishes to use on their T2 Mac and thus I think the PR can be simplified to just saying that Manjaro is not supported by anyone rn. It's possible that some other maintainer interested in Manjaro many join us.

docs/distributions/manjaro/home.md
```diff
@@ -1,3 +1,7 @@
-Welcome to the mbp-manjaro wiki!
+# Deprecation Notice
+
+Manjaro support will be deprecated soon. See [Deprecation Plan](https://wiki.t2linux.org/distributions/manjaro/deprecation) for more information.
+
+# Welcome to the mbp-manjaro wiki!
```
AdityaGarg8, line 5: We could do T2 Manjaro instead of mbp-manjaro

AdityaGarg8 (changes requested):

NoaHimesaka1873:
@AdityaGarg8

AdityaGarg8 (approved):

docs/distributions/manjaro/faq.md
```diff
@@ -1,4 +1,8 @@
-# Migrating from Pykee's old kernel to new LTS kernel
+# Deprecation Notice
+
+Manjaro support will be deprecated soon. See [Deprecation Plan](https://wiki.t2linux.org/distributions/manjaro/deprecation) for more information.
```
Redecorating, line 3: Specifically on T2 macs, not all manjaro

NoaHimesaka1873:
@Redecorating fixed with instruction to download both zip and z01 files.

Redecorating (approved):

PR #334, opened 2022-09-21
Use kekrby's audio config files

AdityaGarg8:


docs/guides/audio-config.md
```diff
@@ -20,33 +20,20 @@ If not present, you'll have to update your bootup kernel params:
...
-- [2019 16" MacBook Pro audio files](https://gist.github.com/kevineinarsson/8e5e92664f97508277fefef1b8015fba) - that laptop has 6 speakers and needs slightly different config
-- [2020 13" MacBook Air audio files](https://gist.github.com/bigbadmonster17/8b670ae29e0b7be2b73887f3f37a057b)
+```bash
+git clone https://github.com/kekrby/t2-better-audio.git
+cd t2-better-audio
+bash ./install.sh
+sudo rm -r $HOME/t2-better-audio
+```
 
 # Using Pipewire instead of Pulseaudio
```
Redecorating, line 32: can this section be removed? Or replaced with just a note on pulseaudio?

networkException:
Does this change anything about state.md?

AdityaGarg8:
I couldn’t get automatic switching between headphones and speakers in state.

Redecorating (commented):
I would like to figure out what needs to be changed to make my internal speakers get mapped correctly before we merge (e.g. atm "front right" is on the left, "front center" is on the right, and "Subwoofer" is static).

docs/guides/audio-config.md
AdityaGarg8: We need this section cause Pulseaudio doesn't work with these?

AdityaGarg8:
> I would like to figure out what needs to be changed to make my internal speakers get mapped correctly before we merge (e.g. atm "front right" is on the left, "front center" is on the right, and "Subwoofer" is static).

Thats strange, cause its mapped correctly for me.

docs/guides/audio-config.md
Redecorating: Oh it's still not working with pulseaudio. Maybe include a link to the previous version of this page for those who don't want to switch?

Redecorating:
> Thats strange, cause its mapped correctly for me.

Hmm I've been checking in the Gnome Settings Audio test thing, does it work properly for you there?
image

AdityaGarg8:
> > Thats strange, cause its mapped correctly for me.
> 
> Hmm I've been checking in the Gnome Settings Audio test thing, does it work properly for you there? image

It did work iirc. I'll still check it again when I get time.

docs/guides/audio-config.md
AdityaGarg8: Well, the previous version won't work as bce is also gonna get patched.

Redecorating: Hmm they'd have to rename the files i guess. I'm not sure if we should break support for pa, would that be a regression?

AdityaGarg8: We could use kekrby's old config files then.

kekrby: [These files](https://github.com/kekrby/t2-better-audio/tree/163ecc6a0dbafbe8107fd5369bb374e343be9c61) should work with `pulseaudio` and an updated `apple-bce`.

kekrby: A `pulseaudio` with [this patch](https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/596.patch) applied should work as well.

kekrby:
I think how to check if `apple-bce` is updated should be included at the introduction part. `sed -n "s/.*\(AppleT2.*\) -.*/\1/p" /proc/asound/cards` can be used to check it as it will output `AppleT2xN` (where `N` is a number) if the driver has been updated and `AppleT2` if it hasn't. 

I can't suggest changes for that part as it was not edited in this PR so can you add this @AdityaGarg8?

kekrby (changes requested):

docs/guides/audio-config.md
```diff
@@ -20,33 +20,20 @@ If not present, you'll have to update your bootup kernel params:
 
 # Audio Configuration Files
 
-In most scenarios, you should use [these files](https://gist.github.com/MCMrARM/c357291e4e5c18894bea10665dcebffb), following the instructions in that gist's `README.md`.
+Simply run the following to set up audio :-
```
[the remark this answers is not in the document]
AdityaGarg8: It was meant to :)

kekrby:
> Does this change anything about state.md?

Audio works well now so I think it can be moved from "partially working" to "working".

AdityaGarg8:
> > Does this change anything about state.md?
> 
> Audio works well now so I think it can be moved from "partially working" to "working".

Is the microphone volume fine for you?

kekrby:
> Is the microphone volume fine for you?

The volume is pretty good for me.

AdityaGarg8:
> > Is the microphone volume fine for you?
> 
> The volume is fine for me.

Still low for me :)

kekrby:
> Still low for me :)
I think you can adjust it in settings, if that is sufficient by itself I don't think it is a big problem. If it requires an external program I guess it should still stay in "partially working" in state.md.

The part about switching to headphones not working should still be removed though.

AdityaGarg8:
> > Still low for me :)
> 
> I think you can adjust it in settings, if that is sufficient by itself I don't think it is a big problem. If it requires an external program I guess it should still stay in "partially working" in state.md.
> 
> The part about switching to headphones not working should still be removed though.

Its written speakers and microphone (though I myself never tried the microphone)

kekrby:
Then only the part about headphone switching should be removed from "not working".

docs/guides/audio-config.md
```diff
@@ -20,33 +20,22 @@ If not present, you'll have to update your bootup kernel params:
...
-- [2019 16" MacBook Pro audio files](https://gist.github.com/kevineinarsson/8e5e92664f97508277fefef1b8015fba) - that laptop has 6 speakers and needs slightly different config
-- [2020 13" MacBook Air audio files](https://gist.github.com/bigbadmonster17/8b670ae29e0b7be2b73887f3f37a057b)
+```bash
+git clone https://github.com/kekrby/t2-better-audio.git
+cd t2-better-audio
+bash ./install.sh
+sudo rm -r $HOME/t2-better-audio
+```
 
 # Using Pipewire instead of Pulseaudio
```
kekrby, line 34:
```suggestion
# Using PulseAudio instead of PipeWire
```

I think it should be this way now since it tells how to use `PulseAudio` instead.

docs/guides/audio-config.md
```diff
@@ -20,33 +20,22 @@ If not present, you'll have to update your bootup kernel params:
 
 # Audio Configuration Files
 
-In most scenarios, you should use [these files](https://gist.github.com/MCMrARM/c357291e4e5c18894bea10665dcebffb), following the instructions in that gist's `README.md`.
+Firstly, check whether you are using an updated apple-bce version by running `sed -n "s/.*\(AppleT2.*\) -.*/\1/p" /proc/asound/cards`. If the output is `AppleT2xN` (where `N` is a number), then this guide can be followed. If its just `AppleT2`, then either update the driver first or follow the [older version](https://github.com/t2linux/wiki/blob/00e882d0d5afb3102aedd02872426dc5fba789a5/docs/guides/audio-config.md) of this guide.
```
kekrby, line 23: I think this it makes more sense to put this in the introduction as the reader is told to check `apple-bce` there.

docs/guides/audio-config.md
```diff
@@ -20,33 +20,22 @@ If not present, you'll have to update your bootup kernel params:
...
-LABEL="pulseaudio_end"
-```
-
-Note: The updated locations for the files in the links above for Pipewire distributions will be:
+Due a bug in PulseAudio, the current configuration does not work with it. A fix has been [already merged](https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/596) but a new version containing the changes hasn't been released yet, so you should use PipeWire instead of PulseAudio unless your distro builds PulseAudio from the main branch or applies the patch before building.
 
-- /usr/share/alsa/cards/AppleT2.conf
-- **/usr/share/alsa-card-profile/mixer/profile-sets/apple-t2.conf**
-- /usr/lib/udev/rules.d/91-pulseaudio-custom.rules
+Different distros have different ways to set up Pipewire. An example to get Pipewire working on Ubuntu, which uses Pulseaudio by default is given [here](https://linuxconfig.org/how-to-install-pipewire-on-ubuntu-linux).
```
kekrby, line 38: I think it should be mentioned earlier that currently only `pipewire` works (without patching) and this should be put after that.

kekrby (changes requested):

docs/guides/audio-config.md
AdityaGarg8: Could you PR the changes you think?

AdityaGarg8:
> > Thats strange, cause its mapped correctly for me.
> 
> Hmm I've been checking in the Gnome Settings Audio test thing, does it work properly for you there? image

@Redecorating , the mapping indeed is wrong for me as well.

docs/guides/audio-config.md
kekrby: I opened #335 for the changes

AdityaGarg8:
So any other changes required?

Redecorating (changes requested):
I don't want to merge this until we either figure out how to fix the incorrect speaker mapping this introduces for 6 speaker macbooks, or we provide instructions on how to use the previous config files for 6 speaker makbooks so people won't have a regression in terms of the audio now being mapped wrong.

Other than fixing or providing advice to workaround that mapping issue for the 16 inch macbooks, it looks good.

kekrby (commented):
I think this shouldn't get merged until the channel map for 6 speaker models is fixed and the `apple-bce` changes are merged to `t2linux/apple-bce-drv`.

AdityaGarg8:
Let's wait for a fix then

kekrby:
I think we can proceed with this now that the channel map issue is solved.

kekrby (changes requested):
I think one last thing that should be done is making the capitalization of letters in PipeWire and PulseAudio consistent. Currently it is a mix of Pulseaudio/Pipewire and PulseAudio/PipeWire.

docs/guides/audio-config.md
```diff
@@ -20,33 +22,18 @@ If not present, you'll have to update your bootup kernel params:
...
-LABEL="pulseaudio_end"
+```bash
+git clone https://github.com/kekrby/t2-better-audio.git
+cd t2-better-audio
+bash ./install.sh
+sudo rm -r $HOME/t2-better-audio
 ```
 
-Note: The updated locations for the files in the links above for Pipewire distributions will be:
+If your distro uses PulseAudio by default, consider switching to PipeWire as the latest stable release of PulseAudio has a bug which prevents audio from working properly. A fix for the bug has been [already merged](https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/596) but a new version containing the changes hasn't been released yet, so you should use PipeWire instead of PulseAudio unless your distro builds PulseAudio from the main branch or applies the patch before building.
```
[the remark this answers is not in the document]
AdityaGarg8: I am having issues with pulse audio

AdityaGarg8:
@Redecorating , could you merge it, if it looks good to you?

AdityaGarg8:
BTW, should there be instructions to remove the old files as well? Or probably write some code to remove the old files in your script @kekrby ?

docs/guides/audio-config.md
```diff
@@ -20,33 +22,18 @@ If not present, you'll have to update your bootup kernel params:
...
-+SUBSYSTEMS=="pci", ATTRS{vendor}=="0x106b", ATTRS{device}=="0x1803", ENV{PULSE_PROFILE_SET}="apple-t2.conf", ENV{ACP_PROFILE_SET}="apple-t2.conf"
+Simply run the following to set up audio :-
 
-LABEL="pulseaudio_end"
+```bash
+git clone https://github.com/kekrby/t2-better-audio.git
+cd t2-better-audio
+bash ./install.sh
+sudo rm -r $HOME/t2-better-audio
 ```
```
kekrby, line 33: These commands should delete the old files. 

Could you please wrap them in triple backticks? I didn't do it as it seemed to break the suggestion UI.

```suggestion

You should also run these commands to delete the old configuration files if you have already set up audio the old way:
sudo rm /usr/share/alsa/cards/AppleT2.conf
sudo rm /usr/lib/udev/rules.d/91-pulseaudio-custom.rules
sudo rm /usr/share/pulseaudio/alsa-mixer/profile-sets/apple-t2.conf

If you use PipeWire, run this command instead of the last one:
sudo rm /usr/share/alsa-card-profile/mixer/profile-sets/apple-t2.conf

```

kekrby (changes requested):

kekrby:
By the way now that there are a lot of commits that this PR includes, I think it makes sense to squash the related ones.

In my opinion 101ccca469b0ebed2998b187ee3763f95dfcc7ef should stay untouched, 0522c1ed3ed6f4f041a2b44220d81a28f38c1642 and 627dadc5c44ba476fcab60b5625e9a6ff3e70ebd should be squashed into one commit and the rest should be squashed into another commit as well.

docs/guides/audio-config.md
AdityaGarg8: https://github.com/kekrby/t2-better-audio/pull/2

Check this out

docs/guides/audio-config.md
```diff
@@ -20,33 +22,18 @@ If not present, you'll have to update your bootup kernel params:
...
-
--SUBSYSTEMS=="pci", ATTRS{vendor}=="0x106b", ATTRS{device}=="0x1803", ENV{PULSE_PROFILE_SET}="apple-t2.conf"
-+SUBSYSTEMS=="pci", ATTRS{vendor}=="0x106b", ATTRS{device}=="0x1803", ENV{PULSE_PROFILE_SET}="apple-t2.conf", ENV{ACP_PROFILE_SET}="apple-t2.conf"
+Simply run the following to set up audio :-
 
-LABEL="pulseaudio_end"
+```bash
+git clone https://github.com/kekrby/t2-better-audio.git
+cd t2-better-audio
+bash ./install.sh
```
kekrby, line 30: There is already a shebang line so I don't think this is necessary.

```suggestion
./install.sh
```

kekrby (changes requested):

AdityaGarg8:
> By the way now that there are a lot of commits that this PR includes, I think it makes sense to squash the related ones.
> 
> In my opinion [101ccca](https://github.com/t2linux/wiki/commit/101ccca469b0ebed2998b187ee3763f95dfcc7ef) should stay untouched, [0522c1e](https://github.com/t2linux/wiki/commit/0522c1ed3ed6f4f041a2b44220d81a28f38c1642) and [627dadc](https://github.com/t2linux/wiki/commit/627dadc5c44ba476fcab60b5625e9a6ff3e70ebd) should be squashed into one commit and the rest should be squashed into another commit as well.

Well as a general trend, I guess we shall be squash and merging the PR.

docs/guides/audio-config.md
```diff
@@ -20,33 +22,18 @@ If not present, you'll have to update your bootup kernel params:
...
-
--SUBSYSTEMS=="pci", ATTRS{vendor}=="0x106b", ATTRS{device}=="0x1803", ENV{PULSE_PROFILE_SET}="apple-t2.conf"
-+SUBSYSTEMS=="pci", ATTRS{vendor}=="0x106b", ATTRS{device}=="0x1803", ENV{PULSE_PROFILE_SET}="apple-t2.conf", ENV{ACP_PROFILE_SET}="apple-t2.conf"
+Simply run the following to set up audio :-
 
-LABEL="pulseaudio_end"
+```bash
+git clone https://github.com/kekrby/t2-better-audio.git
+cd t2-better-audio
+./install.sh
```
Redecorating, line 30: does this need sudo?

docs/guides/audio-config.md
```diff
@@ -20,33 +22,18 @@ If not present, you'll have to update your bootup kernel params:
...
+sudo rm -r $HOME/t2-better-audio
```
Redecorating, line 31: it was cloned without sudo so why is sudo needed to remove it? also this assumes the user was in their home directory when they cloned the repo

Redecorating (changes requested):

docs/guides/audio-config.md
AdityaGarg8: There is a strange issue on Ubuntu that causes error stating something like removing read only files if removing something from $HOME directory.

Redecorating: hmm maybe it'd be simpler to clone it into /tmp? or use rm -rf?

AdityaGarg8: 411234d872de947f2792fa3d23e0d26811fdc8eb

Also, I was thinking to attach the links of the apt repo as an easy alternative method.

AdityaGarg8: You probably could also add the link for your Pacman repo?

docs/guides/audio-config.md
Redecorating: it doesn't https://github.com/kekrby/t2-better-audio/blob/2d835c6e3d4fb0406d2933638d380c8b7fb92700/install.sh#L8

Redecorating (approved):

docs/guides/audio-config.md
AdityaGarg8: Wdym?

Redecorating: I mistakenly thought/remembered the script would say "please run me with sudo" and exit 1, but it doesn't do that so it's not an issue

docs/guides/audio-config.md
kekrby: I think it should be mentioned that this guide should be followed only if the distro being used does not package the files as installing them as a distro package would allow updates to happen just like any other package but manual installation would not.

AdityaGarg8: I guess we can make this change in a separate PR since I've to work on the apt repo a bit more.

AdityaGarg8: I'll merge this if it's good apart from this.

Any other suggestions @Redecorating @kekrby?

kekrby (approved):
LGTM

docs/guides/audio-config.md
```diff
@@ -20,33 +22,18 @@ If not present, you'll have to update your bootup kernel params:
...
-ACTION!="change", GOTO="pulseaudio_end"
-KERNEL!="card*", GOTO="pulseaudio_end"
-
--SUBSYSTEMS=="pci", ATTRS{vendor}=="0x106b", ATTRS{device}=="0x1803", ENV{PULSE_PROFILE_SET}="apple-t2.conf"
-+SUBSYSTEMS=="pci", ATTRS{vendor}=="0x106b", ATTRS{device}=="0x1803", ENV{PULSE_PROFILE_SET}="apple-t2.conf", ENV{ACP_PROFILE_SET}="apple-t2.conf"
+Simply run the following to set up audio :-
 
-LABEL="pulseaudio_end"
+```bash
+sudo git clone https://github.com/kekrby/t2-better-audio.git /tmp/t2-better-audio
```
alex-007, line 28: I don't think you should be a root to write something to /tmp. Isn't it better to not use `sudo` here?

sharpenedblade: For certain distros it is necessary.

PR #338, opened 2022-09-30
Remove unneeded FAQ.md from Arch and update installation guide to suit new ISO release

NoaHimesaka1873:
Now with less instructions!


docs/distributions/arch/faq.md
```diff
@@ -1,17 +0,0 @@
-# Updating Kernel
```
Redecorating, line 1: this will break a link in https://wiki.t2linux.org/guides/wifi-bluetooth/#ensure-kernel-supports-otp-firmware-selection

Redecorating (changes requested):

Redecorating:
I'm also not sure if t2strap is needed, from my understanding arch is meant to either be installed manually so the user understands what's in their system, or with archinstall. t2strap seems to automate things users might like to know about.

I'm thinking an addon to archinstall would be preferable, as in that case users expect things to be automated.

NoaHimesaka1873:
> I'm also not sure if t2strap is needed, from my understanding arch is meant to either be installed manually so the user understands what's in their system, or with archinstall. t2strap seems to automate things users might like to know about.
> 
> I'm thinking an addon to archinstall would be preferable, as in that case users expect things to be automated.

T2strap does three things: omit unneeded/non-working packages (e.g. normal Arch kernel), Installs firmware and audio config, and adds T2 repositories, and selects kernel that's compatible.
The only thing that makes it easier than normal PC arch install is kernel selection. T2strap is syntax-compatible with pacstrap intentionally to give that "manual install" experience. You still need to install other packages. It just gives it feature parity + nicer kernel compared to normal pacstrap. Other than kernel selection, those things are not needed in normal PC arch install.

Also it reduces burden on support channel, like "I forgot to add T2 repo and I can't update!" or "I can't use Wi-Fi/Bluetooth!"

Redecorating:
> T2strap does three things: omit unneeded/non-working packages (e.g. normal Arch kernel), Installs firmware and audio config, and adds T2 repositories, and selects kernel that's compatible.
> The only thing that makes it easier than normal PC arch install is kernel selection. T2strap is syntax-compatible with pacstrap intentionally to give that "manual install" experience. You still need to install other packages. It just gives it feature parity + nicer kernel compared to normal pacstrap. Other than kernel selection, those things are not needed in normal PC arch install.

Can we make it be very verbose and ask for confirmation when it adds or removes packages? e.g. someone might want to install on an external disk have linux-t2 for when booting on the macbook, and linux-zen for when booting their desktop PC, at the moment it looks like it'll silently remove linux-zen without informing the user. Also do the denylist and allowed kernels list need to be not inside the python script?

> Also it reduces burden on support channel, like "I forgot to add T2 repo and I can't update!" or "I can't use Wi-Fi/Bluetooth!"

Looking at the arch channel it seems like there hasn't been much of those issues, it seems like most people haven't been missing adding the repo and getting the packages listed.

An alternative way to ensure people get the right packages would be to have a metapackage like "base-t2" https://github.com/Redecorating/archlinux-t2-packages/blob/metapackages/base-t2/PKGBUILD, this will be able to ensure users have the correct packages, and will let us update it to add new dependencies that users will be prompted to install.

NoaHimesaka1873:
> > T2strap does three things: omit unneeded/non-working packages (e.g. normal Arch kernel), Installs firmware and audio config, and adds T2 repositories, and selects kernel that's compatible.
> > The only thing that makes it easier than normal PC arch install is kernel selection. T2strap is syntax-compatible with pacstrap intentionally to give that "manual install" experience. You still need to install other packages. It just gives it feature parity + nicer kernel compared to normal pacstrap. Other than kernel selection, those things are not needed in normal PC arch install.
> 
> Can we make it be very verbose and ask for confirmation when it adds or removes packages? e.g. someone might want to install on an external disk have linux-t2 for when booting on the macbook, and linux-zen for when booting their desktop PC, at the moment it looks like it'll silently remove linux-zen without informing the user. Also do the denylist and allowed kernels list need to be not inside the python script?
> 
> > Also it reduces burden on support channel, like "I forgot to add T2 repo and I can't update!" or "I can't use Wi-Fi/Bluetooth!"
> 
> Looking at the arch channel it seems like there hasn't been much of those issues, it seems like most people haven't been missing adding the repo and getting the packages listed.
> 
> An alternative way to ensure people get the right packages would be to have a metapackage like "base-t2" https://github.com/Redecorating/archlinux-t2-packages/blob/metapackages/base-t2/PKGBUILD, this will be able to ensure users have the correct packages, and will let us update it to add new dependencies that users will be prompted to install.

For removing package, I'll make it verbose. But for adding packages, since they're all necessary packages to make it properly run, it would be better to leave it silent. Reason why they aren't in script is so I can dynamically update what kernel is allowed. I have a plan to add linux-zen-t2 and linux-lts-t2 near future so it can be added without rebuilding the whole ISO. I didn't consider metapackage because there's less flexibility to make user choose kernel.

Redecorating:
>  I didn't consider metapackage because there's less flexibility to make user choose kernel.

We can have several kernel packages with `provides('linux' 'linux-t2')`, and then pacman will prompt users to select one as `linux-t2` is a dependency of `base-t2`

I'm still not too sure if this is needed for arch (In a way it's an extra deviation from the upstream arch install guide?)  so I'd like to get someone else's opinion too

sharpenedblade:
Normal arch has multiple kernels that provide `linux`, but they are otherwise separate. The way it is supposed to work is that you install the `linux` package and pacman asks you which kernel you want. Also, you could add `replaces=(linux)` to linux-t2, so installing it will remove the normal kernel. I think it is better to just follow the way arch packages its normal kernels.

NoaHimesaka1873:
> Also, you could add `replaces=(linux)` to linux-t2, so installing it will remove the normal kernel. I think it is better to just follow the way arch packages its normal kernels.
This should *never* be done. There might be a case where someone would share one external disk with T2 and non-T2 device.

NoaHimesaka1873:
Superceded by PR #350.

PR #341, opened 2022-10-05
Add repos to DKMS guide + fix headers

AdityaGarg8:


docs/guides/dkms.md
```diff
@@ -2,7 +2,9 @@
 
 This page explains how to install the kernel modules for the Keyboard, Audio, Touchbar and the Ambient Light sensor with DKMS. You will need a patched kernel.
 
-# Do you need to do this?
+## Do you need to do this?
+
+Since many T2 distro maintainers embed these drivers into the kernel itself, you probabaly wouldn't need to install the modules manually. Although it can be used if you are using a kernel other than a T2 kernel provided by the maintainers here.
```
Redecorating, line 7: Maybe say "ship these extra modules alongside the kernel" or "package these modules with the kernel", because embedding them in the kernel might suggest they are compiled with y instead of m for module in the config?

Redecorating (approved):
otherwise lgtm

PR #347, opened 2022-10-30
Add script to seperate out Linux EFI partition

AdityaGarg8:


docs/tools/efi.sh
```diff
@@ -0,0 +1,83 @@
...
+	echo "Windows installation detected. Restoring Windows Bootloader."
+	mkdir /boot/efi/EFI/Boot
+	cp /boot/efi/EFI/Microsoft/Boot/bootmgfw.efi /boot/efi/EFI/Boot/bootx64.efi
+fi
+
+# Update fstab
+
+echo "Updating /etc/fstab."
+OLDUUID=$(lsblk -o mountpoints,uuid | grep -w "/boot/efi" | xargs | cut -d " " -f 2)
+sed -i "s/${OLDUUID}/${NEWUUID}/g" /etc/fstab
```
Redecorating, line 73: Can we exit with an error before mounting partitions and moving files if OLDUUID isn't found in fstab?

docs/tools/efi.sh
```diff
@@ -0,0 +1,83 @@
...
+mkdir /tmp/newefi
+MOUNTPOINT=/$(lsblk -o label,mountpoints | grep -w "$1" | cut -d "/" -f 2-)
+NEWUUID=$(lsblk -o uuid,label | grep -w "$1" | xargs | cut -d " " -f 1)
+umount "${MOUNTPOINT}" 2>/dev/null || true
+mount UUID="${NEWUUID}" /tmp/newefi
+
+# Copy files
+
+echo "Copying Bootloader to new EFI partition."
+cp -r /boot/efi/EFI /tmp/newefi
```
Redecorating, line 45: can we check that there's an efi partition at /boot/efi and fail early if there isn't?

docs/tools/efi.sh
```diff
@@ -0,0 +1,83 @@
...
+if [[ ($input != y) && ($input != Y) ]]
+then
+	echo "Abort!"
+	exit 1
+fi
+
+echo "Changing the EFI partition."
+# Mount the new partition to /tmp/newefi
+
+mkdir /tmp/newefi
```
Redecorating, line 36: maybe confirm that this isn't already there or there's also `mktemp` for making random temporary directories

docs/tools/efi.sh
AdityaGarg8: Check latest push

docs/tools/efi.sh
AdityaGarg8:
```
if [[ ${OLDUUID} != ${FSTABUUID} ]]
then
	echo "Old EFI partition not found in /etc/fstab. Abort!"
	exit 1
fi
```

You should get this error in this case as well.

docs/tools/efi.sh
Redecorating: mktemp makes files and folders with random names so there won't be the issue of if a user has files in /tmp/newefi those files will be deleted.

There's also the option of checking that /tmp/newefi isn't already there.

Redecorating: actually does mkdir fail if tge directory is already there? if it does then this wouldn't be an issue

AdityaGarg8: Yes it fails

AdityaGarg8: I don't think mktemp is required then

Redecorating (approved):

PR #348, opened 2022-10-30
Index: Remove warnings

AdityaGarg8:
These warnings I believe no longer make sense since there has been significant updates to the ISOs and kernels.


networkException (approved):
Yea thats alright

PR #349, opened 2022-10-30
docs/guides/audio-config: recommend rtkit for the best experience

kekrby:


AdityaGarg8:
Could you mention setting up rtkit in a bit more detail?

Like how to set that up, the commands required etc.

PR #352, opened 2022-11-06
amdgpu dpm workaround

lemmyg:
Update Hybrid graphics wiki with some workarounds for amdgpu dynamic power management issues

### DIFF
--- a/docs/guides/hybrid-graphics.md
+++ b/docs/guides/hybrid-graphics.md
@@ -16,7 +16,7 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and 
     Posible workarounds are:
         1. Set iGPU as main gpu.
         2. Set AMD gpu Dynamic Power Management from auto to low.
-        3. To play games to you can set the DPM to high.
+        3. To play games, set the DPM to high.
     
     You test it quickly with: `echo low | sudo tee /sys/bus/pci/drivers/amdgpu/0000:??:??.?/power_dpm_force_performance_level`
 
--- a/docs/guides/hybrid-graphics.md
+++ b/docs/guides/hybrid-graphics.md
@@ -28,7 +28,7 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and 
 
     You can test it quickly with: `echo low | sudo tee /sys/bus/pci/drivers/amdgpu/0000:??:??.?/power_dpm_force_performance_level`
 
-    To apply the low level permanently, create </etc/udev/rules.d/30-amdgpu-pm.rules> file with the following contents:
+    To apply the low level permanently, create `etc/udev/rules.d/30-amdgpu-pm.rules` file with the following contents:
 
     ```plain
     KERNEL=="card0", SUBSYSTEM=="drm", DRIVERS=="amdgpu", ATTR{device/power_dpm_force_performance_level}="low"
--- a/docs/guides/hybrid-graphics.md
+++ b/docs/guides/hybrid-graphics.md
@@ -10,17 +10,21 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and 
 2. If using `DRI_PRIME=1` on programs causes system crashes, with "[CPU CATERR](https://gist.github.com/Redecorating/956a672e6922e285de83fdd7d9982e5e#gistcomment-3719941)" problem reports in macOS, disable dynamic power management with the `amdgpu.dpm=0` kernel argument, or `echo high | sudo tee /sys/bus/pci/drivers/amdgpu/0000:??:??.?/power_dpm_force_performance_level`.
 3. If apple-set-os is loaded, the iGPU will control display brightness, and if the iGPU isn't the boot gpu, the i915 Intel graphics driver will not load, and the display brightness cannot be changed (The exception to this is sometimes when rebooting from macOS Recovery etc, i915 loads fine).
 4. Some users with MacBookPro16,1 and AMD gpu have reported issues:
+
     1. The AMD gpu uses too much power with excessive high temperatures under normal conditions.
     2. The AMD gpu causes system crashes with varied causes including: Shutting down(sudden fan noise and getting T2 chip reset), changing display resolution, having gpu busy when screen is off.
+
     Posible workarounds are:
         1. Set iGPU as main gpu.
         2. Set AMD gpu Dynamic Power Management from auto to low.
         3. To play games, set the DPM to high.
     You can test it quickly with: `echo low | sudo tee /sys/bus/pci/drivers/amdgpu/0000:??:??.?/power_dpm_force_performance_level`
+
     To apply the low level permanently, create </etc/udev/rules.d/30-amdgpu-pm.rules> with the following contents:
         ```plain
         KERNEL=="card0", SUBSYSTEM=="drm", DRIVERS=="amdgpu", ATTR{device/power_dpm_force_performance_level}="low"
         ```
+
     You can also control the AMD gpu DMP with GUI tools such as [radeon-profile](https://github.com/emerge-e-world/radeon-profile).
 
 ## Enabling the iGPU
--- a/docs/guides/hybrid-graphics.md
+++ b/docs/guides/hybrid-graphics.md
@@ -7,28 +7,40 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and 
 ## Issues
 
 1. Resume after suspend is broken, as the GMUX (graphics multiplexer) doesn't connect the iGPU to the display after resuming. For this to be fixed, a Linux driver for `acpi:APP000B:GPUC:` needs to be written (macOS uses AppleMuxControl2.kext). The extra battery life may make this a worthwhile trade-off (about 3 hours to almost 6 hours on a MacBookPro16,1)
+
 2. If using `DRI_PRIME=1` on programs causes system crashes, with "[CPU CATERR](https://gist.github.com/Redecorating/956a672e6922e285de83fdd7d9982e5e#gistcomment-3719941)" problem reports in macOS, disable dynamic power management with the `amdgpu.dpm=0` kernel argument, or `echo high | sudo tee /sys/bus/pci/drivers/amdgpu/0000:??:??.?/power_dpm_force_performance_level`.
+
 3. If apple-set-os is loaded, the iGPU will control display brightness, and if the iGPU isn't the boot gpu, the i915 Intel graphics driver will not load, and the display brightness cannot be changed (The exception to this is sometimes when rebooting from macOS Recovery etc, i915 loads fine).
+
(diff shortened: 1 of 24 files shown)


docs/guides/hybrid-graphics.md
```diff
@@ -9,6 +9,17 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and
 1. Resume after suspend is broken, as the GMUX (graphics multiplexer) doesn't connect the iGPU to the display after resuming. For this to be fixed, a Linux driver for `acpi:APP000B:GPUC:` needs to be written (macOS uses AppleMuxControl2.kext). The extra battery life may make this a worthwhile trade-off (about 3 hours to almost 6 hours on a MacBookPro16,1)
 2. If using `DRI_PRIME=1` on programs causes system crashes, with "[CPU CATERR](https://gist.github.com/Redecorating/956a672e6922e285de83fdd7d9982e5e#gistcomment-3719941)" problem reports in macOS, disable dynamic power management with the `amdgpu.dpm=0` kernel argument, or `echo high | sudo tee /sys/bus/pci/drivers/amdgpu/0000:??:??.?/power_dpm_force_performance_level`.
```
lemmyg, line 10: settings the dpm to low solved this problem for me.

docs/guides/hybrid-graphics.md
```diff
@@ -9,6 +9,17 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and
...
 3. If apple-set-os is loaded, the iGPU will control display brightness, and if the iGPU isn't the boot gpu, the i915 Intel graphics driver will not load, and the display brightness cannot be changed (The exception to this is sometimes when rebooting from macOS Recovery etc, i915 loads fine).
+4. Some users with  MacBookPro16.1 and amdgpu 5500m models have reported problems. Unable to shutdown normally(sudden fan noise), getting T2 chip reset and excessive high temperatures under normal conditions. Posible workarounds are:
+    1. Set amd dynamic power management from auto to low. `echo low | sudo tee /sys/bus/pci/drivers/amdgpu/0000:??:??.?/power_dpm_force_performance_level`
+
+        To apply the low level permanently:
```
Redecorating, line 15: can you put "create `<filename>` with the following contents" at the end of this line, instead of in the code block?

docs/guides/hybrid-graphics.md
```diff
@@ -9,6 +9,17 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and
...
+   
+        ```plain
+            create /etc/udev/rules.d/30-amdgpu-pm.rules
+            KERNEL=="card0", SUBSYSTEM=="drm", DRIVERS=="amdgpu", ATTR{device/power_dpm_force_performance_level}="low"
```
Redecorating, line 19: card0 is not always the amdgpu, sometimes it will be card1, or probably anything up to card9 if you have a bunch of egpu's and have those cards create drm devices first.

Consider using the device, subsystem_vendor, subsystem_device attributes if you want to make it only apply to the internal amd gpu's (for 5300M those values are 7340, 106b, and 0210, respectively, and on 5600M mbp16,4 they would be 7360, 106b, 021e)

docs/guides/hybrid-graphics.md
```diff
@@ -9,6 +9,17 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and
...
 2. If using `DRI_PRIME=1` on programs causes system crashes, with "[CPU CATERR](https://gist.github.com/Redecorating/956a672e6922e285de83fdd7d9982e5e#gistcomment-3719941)" problem reports in macOS, disable dynamic power management with the `amdgpu.dpm=0` kernel argument, or `echo high | sudo tee /sys/bus/pci/drivers/amdgpu/0000:??:??.?/power_dpm_force_performance_level`.
 3. If apple-set-os is loaded, the iGPU will control display brightness, and if the iGPU isn't the boot gpu, the i915 Intel graphics driver will not load, and the display brightness cannot be changed (The exception to this is sometimes when rebooting from macOS Recovery etc, i915 loads fine).
+4. Some users with  MacBookPro16.1 and amdgpu 5500m models have reported problems. Unable to shutdown normally(sudden fan noise), getting T2 chip reset and excessive high temperatures under normal conditions. Posible workarounds are:
```
Redecorating, line 12: MacBookPro16,1 with a comma not a period/dot (see `cat /sys/class/dmi/id/product_name`).

I think it's all MBP16,1 computers that have this sort of issue, not just 5500M (I have had these problems with 5300M).

Additionally the description of the problems is a bit vague here, there are two main things from my understanding:
- The AMD gpu uses too much power and generates too much heat
- The AMD gpu causes system crashes with varied causes including: Shutting down, changing display resolution, having gpu busy when screen is off

Then for the former the solutions are 1: Use iGPU as main gpu 2: set the performance level to low. And for the latter problem, set the performance level to either low or high.

docs/guides/hybrid-graphics.md
```diff
@@ -9,6 +9,17 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and
...
+    1. Set amd dynamic power management from auto to low. `echo low | sudo tee /sys/bus/pci/drivers/amdgpu/0000:??:??.?/power_dpm_force_performance_level`
+
+        To apply the low level permanently:
+   
+        ```plain
+            create /etc/udev/rules.d/30-amdgpu-pm.rules
+            KERNEL=="card0", SUBSYSTEM=="drm", DRIVERS=="amdgpu", ATTR{device/power_dpm_force_performance_level}="low"
+        ```
+    
+    2. To control power_dpm_force_performance_level interactively we can use `https://github.com/emerge-e-world/radeon-profile`
```
Redecorating, line 22: Maybe put this as You can also control the gpu power settings with GUI tools such as \[radeon-profile](link here).

Redecorating (changes requested):
This does duplicate a bit of the stuff in line 10. Maybe we should make a new page and move all the amdgpu related stuff there.

Besides my comments there's also a couple of things the linter picked up so if you could check those when you have time that'd be good too.

docs/guides/hybrid-graphics.md
lemmyg: It would better if we add way to find which card number is the AMD gpu.
It will be easier for users than find out the vendor number.
For example radeon-profile tool uses card0, so it must be a way.

docs/guides/hybrid-graphics.md
lemmyg: ok

lemmyg: about redesigning the page, I think is better if leave that you guys.
You could do it in another revision as you think is better.

lemmyg (commented):
done. let me know if I missing something

docs/guides/hybrid-graphics.md
```diff
@@ -7,9 +7,42 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and
...
+        2. Set AMD gpu Dynamic Power Management from auto to low.
+
+        3. To play games, set the DPM to high.
+
+    You can test it quickly with: `echo low | sudo tee /sys/bus/pci/drivers/amdgpu/0000:??:??.?/power_dpm_force_performance_level`
+
+    To apply the low level permanently, create `etc/udev/rules.d/30-amdgpu-pm.rules` file with the following contents:
+
+    ```plain
+    KERNEL=="card0", SUBSYSTEM=="drm", DRIVERS=="amdgpu", ATTR{device/power_dpm_force_performance_level}="low"
```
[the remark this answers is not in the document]
lemmyg: I don't think it will not work without the kernel. All I'm trying to do is to add some examples.  Then users need to find out what's the best solution.

docs/guides/hybrid-graphics.md
```diff
@@ -7,9 +7,36 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and
...
+
 3. If apple-set-os is loaded, the iGPU will control display brightness, and if the iGPU isn't the boot gpu, the i915 Intel graphics driver will not load, and the display brightness cannot be changed (The exception to this is sometimes when rebooting from macOS Recovery etc, i915 loads fine).
 
+4. Some users with MacBookPro16,1 and AMD gpu have reported issues like AMD gpu uses too much power with excessive high temperatures under normal conditions or sudden high speed fan noise then instant system shut down, getting the T2 chip reset.
+
+    Posible workarounds:
+
+    1. Set iGPU as main gpu.
+
+    2. Set AMD gpu Dynamic Power Management from auto to low.
```
Redecorating, line 21: maybe add "(or high)" here because that's also a way to stop the issue

docs/guides/hybrid-graphics.md
```diff
@@ -7,9 +7,36 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and
...
+
+    You can test it quickly with: `echo low | sudo tee /sys/class/drm/card0/device/power_dpm_force_performance_level`
+
+    To apply the low level permanently, create `etc/udev/rules.d/30-amdgpu-pm.rules` file with the following contents:
```
Redecorating, line 25: **/**etc

docs/guides/hybrid-graphics.md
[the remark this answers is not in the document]
Redecorating: I guess this works for now, but probably mention that the card number could change if you use the igpu or have an egpu. When I have time I might see if it can be done without relying on a consistent card number.

Redecorating (commented):
after the two mentioned changes, it should be good. In future we might see if we can split the amdgpu stuff into a seperate page

docs/guides/hybrid-graphics.md
lemmyg: I would prefer to avoid  high as can lead to thermal issues. In my case I have low by the default and only high when playing games. Some times high can lead to crashes. I found low as most safer setting if you are with amdgpu.

lemmyg: ok Im going to add change

docs/guides/hybrid-graphics.md
lemmyg: this is just an example and prefer to lead users to use low.

lemmyg: all change, please have a look

Redecorating: it seems like it's still a relative path instead of an absolute path (starting with "/"). I think this means it currently only will work if the current directory is /

lemmyg: good catch. done

lemmyg: sorry I didnt understand the **/**etc. I thought you were saying to add comment about high too. my mad

Redecorating: no worries, in hindsight i can see how that could be misinterpreted

Redecorating (approved):

PR #354, opened 2022-11-07
Instructions for rmmod_tb.sh script on Gentoo openrc with elogind

vi-improved:
No idea if this is needed but it may be a good idea to add since Gentoo is a supported distro.

### DIFF
--- a/docs/guides/dkms.md
+++ b/docs/guides/dkms.md
@@ -124,6 +124,7 @@ If you still face an issue, mention it [here](https://github.com/t2linux/wiki/is
 ## Fixing Suspend
 
 Copy [this script](../tools/rmmod_tb.sh) to `/lib/systemd/system-sleep/rmmod_tb.sh`
+If you are using Gentoo with openrc, instead copy the script to `/lib64/elogind/system-sleep/rmmod_tb.sh`
 
 Now run :-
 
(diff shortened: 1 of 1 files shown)


docs/guides/dkms.md
```diff
@@ -125,13 +125,17 @@ If you still face an issue, mention it [here](https://github.com/t2linux/wiki/is
...
+If you are using Gentoo with openrc, instead copy the script to `/lib64/elogind/system-sleep/rmmod_tb.sh`
+
 Now run :-
 
 ```sh
 sudo chmod 755 /lib/systemd/system-sleep/rmmod_tb.sh
 sudo chown root:root /lib/systemd/system-sleep/rmmod_tb.sh
 ```
 
+Change the path to `/lib64/elogind/system-sleep/rmmod_tb.sh` if using openrc on Gentoo as mentioned previously.
```
EmreTech, line 137: Isn't this implied from the previous line you added?

EmreTech (commented):
Just one thing otherwise looks good to me 👍

docs/guides/dkms.md
AdityaGarg8: I guess what @vi-improved did is good enough in this case.

PR #366, opened 2022-12-17
Replace DKMS guide with a new basic set up guide.

AdityaGarg8:


AdityaGarg8:
@Redecorating, can you check this out. The pr is for the preinstall branch btw to ensure that things are synced.

docs/guides/setting-up.md
```diff
@@ -0,0 +1,118 @@
...
+```sh
+echo apple-bce | sudo tee /etc/modules-load.d/t2.conf
+```
+
+## Make modules load on early boot
+
+Having the `apple-bce` module loaded early allows the use of the keyboard for decrypting encrypted volumes (LUKS).
+It also is useful when boot doesn't work, and the keyboard is required for debugging.
+To do this, one must ensure the `apple-bce` module *as well as its dependent modules* are included in the initial ram disk.
+If your distro uses `initramfs-tools` (all debian-based distros), then `/etc/initramfs-tools/modules` stores a list of extra modules to be included and loaded at early boot time.
```
Redecorating, line 54: for this section maybe we can do it as a list?

- on systems with initramfs-tools
    1. step 1
    2. step 2
- systems with mkinitcpio
    1. step 1
    2. step 2
- systems with other initramfs/initrd generation systems
    1. refer to the documentation and ensure the kernel module apple-bce is loaded early

or something like that.

also note that not all arch systems use mkinitcpio, some users may choose to use booster or other ones and will have to consult the respective documentation

docs/guides/setting-up.md
```diff
@@ -0,0 +1,118 @@
...
+   ```
+  
+- Boot into the [macOS Recovery](https://support.apple.com/en-gb/HT201314) and then restart into Linux.
+- Unplug all the external USB keyboards and mouse and then restart into Linux, keeping them unplugged.
+
+If you still face an issue, mention it [here](https://github.com/t2linux/wiki/issues) or on the discord.
+
+## Fixing Suspend
+
+Copy [this script](../tools/rmmod_tb.sh) to `/lib/systemd/system-sleep/rmmod_tb.sh`
```
Redecorating, line 99: if the script is not already there I suppose?

docs/guides/setting-up.md
```diff
@@ -0,0 +1,118 @@
...
+Now run :-
+
+```sh
+sudo chmod 755 /lib/systemd/system-sleep/rmmod_tb.sh
+sudo chown root:root /lib/systemd/system-sleep/rmmod_tb.sh
+```
+
+Change the path to `/lib64/elogind/system-sleep/rmmod_tb.sh` if using OpenRC on Gentoo as mentioned previously.
+
+It unloads the Touchbar modules as they can cause issues for suspend.
```
Redecorating, line 112: and reloads them after resume

docs/roadmap.md
```diff
@@ -16,8 +16,9 @@ Still in case something goes wrong we are not responsible for it, we will try ou
 
 While its technically possible to remove macOS entirely, its strongly encouraged to dual boot it with Linux because :-
 
-1. It acts as a backup in case something goes wrong.
-2. macOS updates often bring along certain firmware updates, which tend to be useful for Linux as well.
+1. To set up Wi-Fi on Linux, you will need Wi-Fi firmware, which can be obtained only from macOS.
```
Redecorating, line 19: maybe say "can be only legally obtained"? to provide a little context as to why the files can't be hosted on a website

mkdocs.yml
```diff
@@ -50,8 +51,8 @@ nav:
           - Notable Contributors: notable-contributors.md
     - Guides:
           - Pre-Install: guides/preinstall.md
+          - Basic setup: guides/setting-up.md
```
Redecorating, line 54: possibly consider "postinstall" as the name, but "setting up" is fine too

docs/guides/setting-up.md
AdityaGarg8: Wdym?

Redecorating: iirc, some t2 specific iso's put that script there already? although it might be an outdated copy of it

AdityaGarg8: I don’t think it causes any functional change if the user follows it again. Moreover the guide initially has a Do I need to follow this section.

mkdocs.yml
AdityaGarg8: Done

docs/roadmap.md
AdityaGarg8: Done

docs/guides/setting-up.md
AdityaGarg8: Done

docs/guides/setting-up.md
AdityaGarg8: Done

AdityaGarg8:
@Redecorating , check now.

docs/guides/postinstall.md
```diff
@@ -0,0 +1,129 @@
...
+
+```sh
+echo apple-bce | sudo tee /etc/modules-load.d/t2.conf
+```
+
+## Make modules load on early boot
+
+Having the `apple-bce` module loaded early allows the use of the keyboard for decrypting encrypted volumes (LUKS).
+It also is useful when boot doesn't work, and the keyboard is required for debugging.
+To do this, one must ensure the `apple-bce` module *as well as its dependent modules* are included in the initial ram disk.
```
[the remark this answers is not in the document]
AdityaGarg8: I'll merge this first to the preinstall branch. Rest editing can be done there.

PR #381, opened 2023-01-17
Remove outdated information from Fedora faq.

sharpenedblade:
The sound and kernel update information are outdated, it works ootb now.

### DIFF
--- a/docs/guides/wifi-bluetooth.md
+++ b/docs/guides/wifi-bluetooth.md
@@ -9,7 +9,7 @@ Check if this command outputs any lines: `modinfo brcmfmac | grep 4387` If it do
 Refer to the "Updating Kernel" section on your distro's FAQ for instructions if you need to update your kernel:
 
 - [Arch](https://wiki.t2linux.org/distributions/arch/faq/#updating-kernel)
-- [Fedora](https://docs.fedoraproject.org/en-US/quick-docs/securing-the-system-by-keeping-it-up-to-date/#manual-updating-using-gui)
+- [Fedora](https://github.com/mikeeq/mbp-fedora-kernel#how-to-update-mbp-fedora-kernel)
 - [Manjaro](https://wiki.t2linux.org/distributions/manjaro/faq/#updating-kernel)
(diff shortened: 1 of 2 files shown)


AdityaGarg8:
Can you explain the kernel update change you have done here.

sharpenedblade:
We used to need to run a script, but now it is the same as a normal upgrade with `dnf`, so I linked to fedoras docs on how to do that.

docs/guides/wifi-bluetooth.md
```diff
@@ -9,7 +9,7 @@ Check if this command outputs any lines: `modinfo brcmfmac | grep 4387` If it do
 Refer to the "Updating Kernel" section on your distro's FAQ for instructions if you need to update your kernel:
 
 - [Arch](https://wiki.t2linux.org/distributions/arch/faq/#updating-kernel)
-- [Fedora](https://wiki.t2linux.org/distributions/fedora/faq/#updating-kernel)
+- [Fedora](https://docs.fedoraproject.org/en-US/quick-docs/securing-the-system-by-keeping-it-up-to-date/#manual-updating-using-gui)
```
AdityaGarg8, line 12: As per mikeeq’s repo instructions, the def repo has been introduced in fedora 37, and I believe that no user of Fedora 37 would have to follow these instructions since all releases of the same have Asahi Linux patches.

If a person is in a situation that he had to follow the “Updating Kernel” section, it means that they are using some old release, and as per mikeeq’s repo, the dnf way shouldn’t work for them right?


So I guess we should keep it as such? The faq changes look fine though.

AdityaGarg8 (changes requested):

docs/guides/wifi-bluetooth.md
sharpenedblade: > If a person is in a situation that he had to follow the “Updating Kernel” section, it means that they are using some old release, and as per mikeeq’s repo, the dnf way shouldn’t work for them right?

I didnt think of it that way, I linked it to mikeeqs instructions now.

AdityaGarg8 (approved):

PR #384, opened 2023-01-27
Add blendOS

NoaHimesaka1873:


AdityaGarg8:
Due to bug in the installer, we need to put this on Hold.

AdityaGarg8:
Also, better add a documentation to install the OS. The automatic partitioning method will erase off macOS and users may follow that.

AdityaGarg8:
As a second thought, it’s not possible to document every distro, so imo https://github.com/t2linux/wiki/commit/a9a365350755b78188c7b4daf2e15b8710212b12 should be good.

Although I’ve pushed it, and am merging this PR, still @Redecorating if you think we can change something, them do tell me.

PR #392, opened 2023-02-07
Update copyright date

thefirethirteen:
It's outdated by 2 years by now.

Note that it is also outdated on https://t2linux.org and I don't know where to contribute to the site itself.


networkException:
Thanks!

PR #394, opened 2023-02-10
hybrid-graphics: wip update to work with new gmux patches

Redecorating:
depends on https://github.com/t2linux/linux-t2-patches/pull/5

### DIFF
--- a/docs/guides/hybrid-graphics.md
+++ b/docs/guides/hybrid-graphics.md
@@ -93,13 +93,13 @@ Currently the Radeon 5600M AMD GPU on MacBookPro16,4 is [not working](https://lo
 
     2. Install apple-os-set loader from [here](https://github.com/Redecorating/apple_set_os-loader) using macOS, and put it in your Linux EFI partition.
 
-3. Restart into macOS Recovery by immediately pressing and holding Command+R on startup.
+2. Restart into macOS Recovery by immediately pressing and holding Command+R on startup.
 
     1. Open the terminal there and run `nvram fa4ce28d-b62f-4c99-9cc3-6815686e30f9:gpu-power-prefs=%01%00%00%00`.
 
-4. Restart into Linux. You should now be able to access your Linux installation.
+3. Restart into Linux. You should now be able to access your Linux installation.
 
-5. Follow the instructions [above](#enabling-the-igpu).
+4. Follow the instructions [above](#enabling-the-igpu).
 
 ## Use on Windows
 
--- a/docs/guides/hybrid-graphics.md
+++ b/docs/guides/hybrid-graphics.md
@@ -1,44 +1,47 @@
 # Introduction
 
-This page describes how to use the iGPU on MacBookPro's with Hybrid Graphics (2 GPUs). 13 inch MacBooks only have an iGPU, and do not need this. Using the iGPU means you can save power by turning off the more powerful AMD dGPU when you don't need it.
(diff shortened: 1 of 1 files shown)


AdityaGarg8:
In the guide having instructions to compile apple-os-set, maybe we can replace that with an easier way to download the pre compiled one from your repo. Something like:

```
sudo mv /boot/efi/EFI/bootx64.efi /boot/efi/EFI/bootx64_original.efi
wget (or curl) the bootloader (probably provide options for silent version as well here)
```

AdityaGarg8:
Also for MBP 16,4, since now we have gmux, can we simply add the .conf file using a Linux ISO, or a third party Linux driver for macOS instead of following the complicated guide?

networkException:
> maybe we can replace that with an easier way to download the pre compiled one from your repo

compiling that yourself should stay a documented option in any case

Redecorating:
> Also for MBP 16,4, since now we have gmux, can we simply add the .conf file using a Linux ISO, or a third party Linux driver for macOS instead of following the complicated guide?

yeah we should be able to simply that section a lot, i'll figure out how best to do that later.

Redecorating:
> In the guide having instructions to compile apple-os-set, maybe we can replace that with an easier way to download the pre compiled one from your repo. Something like:
> 
> 
> 
> ```
> 
> sudo mv /boot/efi/EFI/bootx64.efi /boot/efi/EFI/bootx64_original.efi
> 
> wget (or curl) the bootloader (probably provide options for silent version as well here)
> 
> ```


for apple-set-os i kinda want to see if we can move to:

systemd-boot: use kernel patch for apple-set-os
grub: chainload some.efi, then boot linux (or get a command added into grub itself)
refind: get its apple-set-os command to work

then systemd-boot will just work, grub will just require installation of a package that adds grub config to run that efi file, and refind will just need the user to edit their config

AdityaGarg8:
It would be nice if we could get this merged as well since Linux 6.2 is out

Redecorating:
I don't think we have any kernels released yet that support it.

AdityaGarg8:
Does the main branch of linux-t2-patches support this?

AdityaGarg8:
Cause Ubuntu is goona release a Linux 6.2 soon

docs/guides/hybrid-graphics.md
```diff
@@ -72,83 +69,43 @@ This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and
...
 
 ## MacBookPro16,4
 
-The AMD GPU on MacBookPro16,4 is not yet compatible with Linux. As a workaround :-
+Currently the Radeon 5600M AMD GPU on MacBookPro16,4 is [not working](https://lore.kernel.org/all/3AFB9142-2BD0-46F9-AEA9-C9C5D13E68E6@live.com/) with Linux. As a workaround :-
 
 ### If you are able to edit your kernel command line :-
 
-1. This workaround recommends GRUB as the bootloader. If you want to use some other bootloader, you need to figure out how to get an equivalent of step 4 and 6 working with your bootloader.
+1. Edit the kernel command line of this boot and add the `nomodeset` kernel parameter. This will enable you to access your Linux system in safe graphics.
```
AdityaGarg8, line 82: We could use `modprobe apple-gmux force_igd=y` instead of `nomodeset`?

AdityaGarg8 (changes requested):

docs/guides/hybrid-graphics.md
Redecorating: That won't work if apple-set-os isn't active yet.

AdityaGarg8: Makes sense

PR #407, opened 2023-03-18
feat: add touchbard as an option to control touch bar

NoaHimesaka1873:
Yes, it's complete enough to be suggested to users, I think.

### DIFF
--- a/docs/guides/postinstall.md
+++ b/docs/guides/postinstall.md
@@ -91,6 +91,8 @@ The steps to be followed vary depending upon the initramfs module loading mechan
 
 The Touch Bar can be set up by running [this script](../tools/touchbar.sh) **in Linux** using `bash /path/to/script`. Make sure your Linux kernel and macOS is updated before running this script.
 
+If you're using Arch Linux or EndeavourOS, you have an option to install `touchbard` package and use that instead. For more information, visit [touchbard repository](https://github.com/NoaHimesaka1873/touchbard).
+
 After running this script, if you wish to change the default mode of the Touch Bar, run `sudo touchbar` and choose the mode you wish.
 
 In case your Touch Bar is unable to change modes on pressing the fn key, you could try the following :-
--- a/docs/guides/postinstall.md
+++ b/docs/guides/postinstall.md
@@ -91,7 +91,7 @@ The steps to be followed vary depending upon the initramfs module loading mechan
 
 The Touch Bar can be set up by running [this script](../tools/touchbar.sh) **in Linux** using `bash /path/to/script`. Make sure your Linux kernel and macOS is updated before running this script.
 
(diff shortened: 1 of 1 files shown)


AdityaGarg8:
I'd like to know the difference between touchbard and the currently in use touchbar script.

AdityaGarg8:
The difference in terms of functionality for the end user

NoaHimesaka1873:
@AdityaGarg8 Changing touchbar mode does not change touchbar mode while booting. Currently touchbar script touches both module option and current touchbar mode.

AdityaGarg8:
But I believe changing the mode while booting is more preferable. Atleast that what’s there in Windows Bootcamp.

NoaHimesaka1873:
Bootcamp doing that doesn't mean we should follow its behavior. It gives user more choice since user can use particular mode one moment and upon reboot it will go back to default mode without manually changing back.

AdityaGarg8:
Say for example I like having f keys by default always. So I’ll have to set them by in the terminal every time I boot Linux?

NoaHimesaka1873:
No, there's config file to set boottime default.

AdityaGarg8:
So users will have to manually edit the config file for the condition I meantioned previously.

NoaHimesaka1873:
Yes. Although I have a plan to implement changing default mode via command line.

AdityaGarg8:
> Yes. Although I have a plan to implement changing default mode via command line.

I believe we should have that implemented that first, cause I don't find a good reason to replace the touchbar script with touchbard.

Although @Redecorating and @kekrby may have separate views.

NoaHimesaka1873:
That's why I placed it as secondary option!

AdityaGarg8:
> That's why I placed it as secondary option!

Probably make it clear then.

docs/guides/postinstall.md
```diff
@@ -91,6 +91,8 @@ The steps to be followed vary depending upon the initramfs module loading mechan
 
 The Touch Bar can be set up by running [this script](../tools/touchbar.sh) **in Linux** using `bash /path/to/script`. Make sure your Linux kernel and macOS is updated before running this script.
 
+As a secondary option, if you want to try something new and you're using Arch Linux or EndeavourOS, you could install `touchbard` package and use that instead. For more information, visit [touchbard repository](https://github.com/NoaHimesaka1873/touchbard).
```
AdityaGarg8, line 94: Isn't it available for Manjaro?

AdityaGarg8 (changes requested):

docs/guides/postinstall.md
NoaHimesaka1873: I didn't upload the package to Manjaro repository. I shall upload and add Manjaro too.

PR #415, opened 2023-05-02
Change fan guide to use t2fand and remove 'secondary option' from touchbard

NoaHimesaka1873:


AdityaGarg8:
Here you should add instructions to add the pacman repo as well.

NoaHimesaka1873:
> Here you should add instructions to add the pacman repo as well.

I didn't add it because I assumed the user would have followed t2linux Arch installation guide, which of course adds the necessary repositories. I'll add warning to indicate that.

AdityaGarg8:
I'd rather separate fan into 3 sections

1st for Arch based distros
2nd for Debian and Ubuntu based distros
3rd for rest cases

1st and 2nd having instructions to add repo and install the package.

NoaHimesaka1873:
Damn linter sucks it can't even properly parse numbered list

docs/guides/fan.md
```diff
@@ -1,38 +1,38 @@
...
+
+        1. Clone the repository by running `git clone https://github.com/NoaHimesaka1873/t2fand`
+
+        2. Change directory into the cloned repository.
+
+        3. Install it by running `make install`.
+
+2. Enable daemon by running `sudo systemctl enable --now t2fand`.
+
+3. Edit the config and restart the daemon if needed.
```
AdityaGarg8, line 33: I do not find any instructions to restart the daemon. Can you elaborate this?

AdityaGarg8 (changes requested):

docs/guides/fan.md
NoaHimesaka1873: Done.

PR #418, opened 2023-05-06
Propose to add hid-apple to mkinitcpio in arch install

strifel:
When using the current config the keyboard drivers were not loaded and, in a setup with encryption, one could therefore only enter the encryption key with an second keyboard

### DIFF
--- a/docs/distributions/arch/installation.md
+++ b/docs/distributions/arch/installation.md
@@ -41,7 +41,7 @@ You will need:
 
 6. Continue following the Arch Wiki's guide until you get to installing a bootloader.
 
-7. Add `apple-bce` to the `MODULES` list in `/etc/mkinitcpio.conf`, and then run `mkinitcpio -P`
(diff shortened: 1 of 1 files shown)


AdityaGarg8:
Isn’t `apple-bce` enough for keyboard. Requesting a review from @Redecorating.

strifel:
> Isn’t `apple-bce` enough for keyboard. 

It did not work for me yesterday on a mbp 2018 16“

Redecorating:
Hmm are you using mkinitcpio? For me its always included enough modules such that the keyboard works in the initramfs stage. I don't mind adding hid-apple to this list - but also please add it to this section as well so we stay consistent. https://wiki.t2linux.org/guides/postinstall/#make-modules-load-on-early-boot

strifel:
Somehow it works now. Thanks!

networkException:
¯\_(ツ)_/¯

PR #420, opened 2023-05-08
Update fedora ISO url to point at new repo

sharpenedblade:
I moved my kernel and ISO repos to the `t2linux-fedora` org, so this updates the link in the wiki page.


AdityaGarg8:
We could move the repos to official t2linux organisation instead? Cc @aunali1

PR #449, opened 2023-08-05
Cleanup duplication

sharpenedblade:
Some steps are also in the post-install guide, and are available on all other supported distros ootb

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -47,7 +47,7 @@ Fedora also has a prebuilt ISO and a graphical installer. Most things work out o
 NixOS has both prebuilt graphical and command line installers. Support for T2 devices is provided in the form of a `nixos-hardware` module.
 Importing the module is the only thing you have to do to get your configuration working on your Mac.
 
-If you wish to use another distribution, you can install it normally, then also follow the steps to [install the kernel yourself manually](https://wiki.t2linux.org/guides/kernel/) and follow the [post installation steps](https://wiki.t2linux.org/roadmap/#configuring-the-installation).
+If you wish to use another distribution, you can install it normally, then follow the steps to [install the kernel yourself manually](https://wiki.t2linux.org/guides/kernel/) and follow the [post installation steps](https://wiki.t2linux.org/roadmap/#configuring-the-installation).
 
 ## Installing
 
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -38,34 +38,30 @@ Arch Linux probably has the most documentation, both officially for the whole pr
 t2linux wiki with a really detailed and up to date install guide. On top of that there is also a work in progress `archinstall` script.
 Keep in mind however that the whole process is done in the command line.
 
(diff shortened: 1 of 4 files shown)


docs/roadmap.md
```diff
@@ -38,34 +38,30 @@ Arch Linux probably has the most documentation, both officially for the whole pr
...
 
-Fedora also has a prebuilt ISO and a graphical installer. Most things work out of the box, though audio and WiFi generally need to be set up as in the guides below. Then they should work flawlessly.
+Fedora also has a prebuilt ISO and a graphical installer. Most things work out of the box.
 
 NixOS has both prebuilt graphical and command line installers. Support for T2 devices is provided in the form of a `nixos-hardware` module.
 Importing the module is the only thing you have to do to get your configuration working on your Mac.
 
-If you wish to use another distribution, you can install it normally, also follow the steps to [install the kernel yourself manually](https://wiki.t2linux.org/guides/kernel/) and follow the [post installation steps](https://wiki.t2linux.org/roadmap/#configuring-the-installation).
+If you wish to use another distribution, you can install it normally, then also follow the steps to [install the kernel yourself manually](https://wiki.t2linux.org/guides/kernel/) and follow the [post installation steps](https://wiki.t2linux.org/roadmap/#configuring-the-installation).
```
AdityaGarg8, line 50: Maybe use “then” instead of “then also”?

docs/roadmap.md
```diff
@@ -38,34 +38,30 @@ Arch Linux probably has the most documentation, both officially for the whole pr
...
 
-Follow the [Pre-installation steps](https://wiki.t2linux.org/guides/preinstall) to prepare your Mac to install Linux and head over to the appropriate guide of the distro of your choice as mentioned in that guide.
+Follow the [pre-installation steps](https://wiki.t2linux.org/guides/preinstall) to prepare your Mac to install Linux and head over to the appropriate guide of the distro of your choice as mentioned in that guide.
 
 ## Configuring the Installation
 
-After successfully booting into your new installation, you will need to configure a few things:
+After successfully booting into your new installation, you will need to follow these steps:
 
--   [Performing a basic setup](https://wiki.t2linux.org/guides/postinstall/)
```
AdityaGarg8, line 62: Why have we removed the postinstall link?

AdityaGarg8 (changes requested):

docs/roadmap.md
sharpenedblade: In the posinstall guide:

> Installing a kernel for T2 support

covered in pre-install guide

> Add necessary kernel paramaters

> Make modules load on boot

> Make modules load on early boot

Enabled by default in Fedora and Ubuntu (maybe other distros too?), it would be better to link from the distros install steps to avoid confusing new users using "user freindly" distros.

> Setting up the Touch Bar

This is useful, I will link to it

> Wi-Fi and Bluetooth

Completed in the previous step on the roadmap

> Network Manager recurrent notifications

Also useful, I will link to it

PR #472, opened 2023-11-21
Update installation.md to init keyring

augustdolan:
We need to initialize the keyring to be able to install in new root. Both `pacstrap` and `t2strap` failed due to this issue. However, I did not test whether `t2strap` would succeed or not with `-K` flag so I didn't add it to the fix.


NoaHimesaka1873:
Not needed, new ISO with new keyring on the way.

PR #479, opened 2023-12-06
fix(eos): instruct correct mount-point for EFI partition

flaw:
Apparently the default has changed from /boot/efi to /efi in Galileo.

(ref: https://discord.com/channels/595304521857630254/942084499381047388/1181616583563558982)


NoaHimesaka1873:
LGTM, merging!

PR #481, opened 2023-12-08
Add instructions for MacOs-like gestures

smileBeda:
Adds a section with 2 gists showing how we can get as close as possible to MacOs gestures on an (Arch) Linux Gnome with Xorg


AdityaGarg8:
You need to address the lint error

smileBeda:
- I hope list indent is now correct. PS: I never saw an MD list with 4 indents. It is always no indent, first level 2 indents, and so on.
- It is a `clickpad` by definition. But I see now that Apple calls it a `trackpad`. Weirdos :P

PR #484, opened 2023-12-10
Update installation.md

### LINKED ISSUE #406: Adding install step for bluetooth on Arch/EndeavourOS
Hi, I've successfully installed EndeavourOS on my Mac Mini 2018 (loving it, thank you so so so much). I wasn't able to use Bluetooth and after help from @Redecorating on Discord, I learned it isn't enabled by default on Arch(-derivatives).

The command `sudo systemctl enable --now bluetooth` made it work without any issue. Would it be worth adding this to the EndeavourOS installation guide?

Thank you all for you hard work with this, it's great to be able to properly use this MacMini, finally!!

### LINKED ISSUE #483: EOS ships with Bluetooth disabled
https://wiki.t2linux.org/guides/wifi-bluetooth/ says:
> If you're running Arch or EndeavourOS and have apple-bcm-firmware installed, you do not need to follow this guide. 

But that is not true, at least the T2 EOS disables Bluetooth to 100% and you need to `systemctl enable --now bluetooth` to even be able to trun on/off bluetooth.
Only then Bluetooth will actually work and allow you to tur it on/off,connect or tether.

Let me know if it is OK to make a PR to add said notice.

NoaHimesaka1873:
Closes #406 and closes #483.

### DIFF
--- a/docs/distributions/endeavouros/installation.md
+++ b/docs/distributions/endeavouros/installation.md
@@ -31,3 +31,5 @@ You will need:
 5. You can follow the [Fan guide](https://wiki.t2linux.org/guides/fan/) after rebooting into your install if your fan isn't working or if you want to customize how/when your fan will run.
 
 6. You now will be able to select your EndeavourOS install in the macOS Startup Manager by holding option at boot.
+
+7. Optionally, enable `bluetooth` daemon if you want to use Bluetooth.


NoaHimesaka1873:
I did not intend to add this part because it was upstream default, but someone *cough* @smileBeda *cough* complained about it, I added it for now.
I do not intend to ship additional tweaks beyond T2 support for EndeavourOS.

PR #488, opened 2023-12-20
Issue: inactive keyboard in initramfs on encrypted partitions

hamidrezaniazi:
## Description:
This pull request addresses an issue related to keyboard availability during the `initramfs` step. If the partition is encrypted, `hid_apple` must be added to the `MODULES` list in `/etc/mkinitcpio.conf` to ensure proper functionality. This step is necessary to prevent the keyboard from becoming unavailable while entering the encryption password.

### DIFF
--- a/docs/distributions/arch/installation.md
+++ b/docs/distributions/arch/installation.md
@@ -42,14 +42,14 @@ You will need:
 
 8. Add `hid_apple` to the `MODULES` list in `/etc/mkinitcpio.conf` only if your partition is encrypted (otherwise you can skip this step), and then run `mkinitcpio -P`
 
-8. Enable `t2fand` and `tiny-dfr` by running:
+9. Enable `t2fand` and `tiny-dfr` by running:
 
    ```bash
    sudo systemctl enable t2fand
    sudo systemctl enable tiny-dfr
    ```
 
-9. Install a bootloader, GRUB is easier, but you can also use systemd-boot. Don't do both.
+10. Install a bootloader, GRUB is easier, but you can also use systemd-boot. Don't do both.
 
     -   Installing Grub:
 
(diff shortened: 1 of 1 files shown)


AdityaGarg8:
information about encryption is mentioned here: https://wiki.t2linux.org/guides/postinstall/#make-modules-load-on-early-boot

Also, is hid_apple really needed on T2 Macs cause apple-bce is the driver that manages keyboard?

sharpenedblade:
`apple-bce` is enough for the keyboard on fedora, I dont think the arch kernel is too diferent

Redecorating:
for arch, do you have the keyboard hook enabled? i think that one will automatically include hid-apple (but unfortunately not apple-bce)

hamidrezaniazi:
> Also, is hid_apple really needed on T2 Macs cause apple-bce is the driver that manages keyboard? @AdityaGarg8

> `apple-bce` is enough for the keyboard on fedora, I dont think the arch kernel is too different @sharpenedblade 

`apple-bce` alone doesn't enable the keyboard between the bootloader and user authentication, specifically when entering the password to unlock the encrypted `volgroup`. However, post-encryption, `apple-bce` indeed takes over for keyboard support. Including both `hid_apple` and `apple-bce` in the modules (`MODULES=(hid_apple apple-bce)`) works well, and I've tested it.

I haven't found an alternative in the wiki for this case. If there's anything I missed or if you have other suggestions, I'd appreciate your insights!

> for arch, do you have the keyboard hook enabled? i think that one will automatically include hid-apple @Redecorating

Yes, the `keyboard` hook was enabled, but it did not automatically include `hid_apple`.

Redecorating:
can you check if you have the keyboard mkinitcpio hook before the encrypt hook? I think it might have to be in that order.

HOOKS=(base udev autodetect modconf kms keyboard keymap consolefont block encrypt filesystems fsck)
is the example order in the arch wiki fwiw

hamidrezaniazi:
> can you check if you have the keyboard mkinitcpio hook before the encrypt hook? I think it might have to be in that order. @Redecorating

Yes, confirming that the order aligns with the mentioned one, `keyboard` is before `encrypt`, so it cannot be the case.

AdityaGarg8:
Request for review from @Redecorating

PR #500, opened 2024-02-14
Improve the linux part of the firmware script

sharpenedblade:
It should work better now, and doesnt mess with mounts


AdityaGarg8:
I actually was thinking of making binaries that run only in macOS using pyinstaller.

It would make a .tar file in macOS itself then.

AdityaGarg8 (changes requested):
We cannot use python3 on macOS since it's not shipped with it. User will have to install python3. So, either ship a binary with macOS or use Linux to rename.

sharpenedblade:
> We cannot use python3 on macOS since it's not shipped with it. User will have to install python3. So, either ship a binary with macOS or use Linux to rename.

`/usr/bin/python3` is from apple for me, it might be xcode dev tools but that can be installed in 1 command (`xcode-select --install`).

AdityaGarg8:
> > We cannot use python3 on macOS since it's not shipped with it. User will have to install python3. So, either ship a binary with macOS or use Linux to rename.
> 
> `/usr/bin/python3` is from apple for me, it might be xcode dev tools but that can be installed in 1 command (`xcode-select --install`).

So we don't want user to install anything as far as macOS is considered.

docs/tools/firmware.sh
```diff
@@ -102,30 +78,13 @@ case "$os" in
 				fi
 			fi
 		done
-		sudo modprobe -r brcmfmac || true
-		sudo modprobe brcmfmac || true
-		sudo modprobe -r hci_bcm4377 || true
-		sudo modprobe hci_bcm4377 || true
-		echo "Cleaning up"
-		sudo rm -r /tmp/apple-wifi-fw
-		echo "Keeping a copy of the firmware and the script in the EFI partition shall allow you to set up Wi-Fi again in the future by running this script or the commands told in the macOS step in Linux only, without the macOS step. Do you want to keep a copy? (y/N)"
```
AdityaGarg8, line 111: Why have the cleanup commands been removed?

AdityaGarg8 (changes requested):

AdityaGarg8:
Also, don't make the user install any additional software on macOS, that includes python3

docs/tools/firmware.sh
sharpenedblade: We dont use `/tmp/apple-wifi-fw` anymore, we mount the ESP to `$mountpoint`.

The firmware tar is pretty tiny (~28Mb), most users can just leave it in the ESP. If something happens to their firmware they can just run `firmware.sh` again instead of rebooting to macos to get the firmware.

Its better to just reboot the system instead of reloading modules since its more stable (at least for me).

sharpenedblade:
> Also, don't make the user install any additional software on macOS, that includes python3

I am going to try to embed python, but that makes the firmware script huge and it needs custom build steps which are very finicky. The python packaging ecosystem is a huge mess, and apple makes it very hard to distribute stuff on macos.

docs/tools/firmware.sh
AdityaGarg8: There should be option to remove the firmware from ESP. You shouldn't enforce this on the user. Also, reloading the modules works for me everytime. Probably give a message that try restarting if wifi still not working.

AdityaGarg8:
> > Also, don't make the user install any additional software on macOS, that includes python3
> 
> I am going to try to embed python, but that makes the firmware script huge and it needs custom build steps which are very finicky. The python packaging ecosystem is a huge mess, and apple makes it very hard to distribute stuff on macos.

Or we could just use Linux to rename. I had said I'll prefer macOS, but since embedding python is difficult, it's better to stay on Linux.

AdityaGarg8:
And if we want to embed, I guess pyinstaller is perfect for this scenario.

docs/tools/firmware.sh
sharpenedblade: > Also, reloading the modules works for me everytime

Actually, the modules are reloaded, its just moved to later in the script.

sharpenedblade:
This just uses new mountpoints now, I will make a new PR with the python script running on macos

docs/tools/firmware.sh
```diff
@@ -43,72 +44,47 @@ case "$os" in
...
+		echo "Installing extra firmware"
 		for file in "$mountpoint/brcmfmac4364b2-pcie.txt" \
-		            "$mountpoint/brcmfmac4364b2-pcie.txcap_blob"
+			"$mountpoint/brcmfmac4364b2-pcie.txcap_blob"
 		do
 			if [ -f "$file" ]
 			then
 				if [[ ${1-default} = -v ]]
 				then
-					sudo cp -v $file /lib/firmware/brcm
```
AdityaGarg8, line 99: Why have you remove sudo from various places like this

AdityaGarg8 (changes requested):

docs/tools/firmware.sh
sharpenedblade: I changed the instructions to run the whole script with sudo, since some sudo configs have a really short timeout.

sharpenedblade: Reverted

docs/tools/firmware.sh
```diff
@@ -43,54 +44,34 @@ case "$os" in
 				cp /usr/share/firmware/wifi/C-4364__s-B2/${txcapblob} "/Volumes/${EFILABEL}/brcmfmac4364b2-pcie.txcap_blob"
 			fi
 		fi
-		echo "Copying this script to EFI"
-		cd - >/dev/null
-		cp "$0" "/Volumes/${EFILABEL}/firmware.sh"|| (echo -e "\nFailed to copy script.\nPlease copy the script manually to the EFI partition using Finder\nMake sure the name of the script is firmware.sh in the EFI partition\n" && echo && read -p "Press enter after you have copied" && echo)
+		echo "Copying this script to the ESP"
+		cp "$0" "/Volumes/${EFILABEL}/firmware.sh"|| (echo -e "\nFailed to copy script.\nPlease copy the script manually to the EFI partition using Finder\n" && echo && read -p "Press enter after you have copied" && echo)
```
AdityaGarg8, line 48: Why did you remove the "make sure the name of the script is firmware.sh" when you have asked the user to run /tmp/apple-wifi-efi/firmware.sh in the following commands.

docs/tools/firmware.sh
```diff
@@ -43,54 +44,34 @@ case "$os" in
...
 		echo "Unmounting the EFI partition"
-		sudo diskutil unmount disk0s1
+		sudo diskutil unmount "/Volumes/${EFILABEL}/"
 		echo
-		echo -e "Run the following commands or run this script itself in Linux now to set up Wi-Fi :-\n\nsudo umount /dev/nvme0n1p1\nsudo mkdir -p /tmp/apple-wifi-efi\nsudo mount /dev/nvme0n1p1 /tmp/apple-wifi-efi\nbash /tmp/apple-wifi-efi/firmware.sh\n"
+		echo -e "Run the following commands or run this script itself in Linux now to set up Wi-Fi:\n\n    sudo mkdir -p /tmp/apple-wifi-efi\n    sudo mount /dev/nvme0n1p1 /tmp/apple-wifi-efi\n    /tmp/apple-wifi-efi/firmware.sh\n"
```
AdityaGarg8, line 52: Add these to the wifi guide as well.

docs/tools/firmware.sh
```diff
@@ -43,54 +44,34 @@ case "$os" in
...
-		fi
-		cd /lib/firmware
-		if [[ ${1-default} = -v ]]
-		then
-			sudo tar xvf /tmp/apple-wifi-fw/firmware.tar
-		else
-			sudo tar xf /tmp/apple-wifi-fw/firmware.tar
-		fi
-
+		echo "Installing extra firmware"
```
AdityaGarg8, line 72: This info is not needed.

docs/tools/firmware.sh
```diff
@@ -124,8 +100,17 @@ case "$os" in
 				fi
 			done
 		fi
-		echo "Running post-installation scripts"
-		exec sudo sh -c "umount /dev/nvme0n1p1 && mount -a && rmdir /tmp/apple-wifi-efi && echo Done!"
+
+		sudo umount $mountpoint
+		sudo rm -r $workdir/*
+		sudo rmdir $mountpoint
+		echo "Reloading Wi-Fi drivers"
```
AdityaGarg8, line 107: Reloading wifi and Bluetooth drivers

AdityaGarg8 (changes requested):

sharpenedblade:
I fixed the requested changes.

AdityaGarg8 (approved):

PR #516, opened 2024-04-11
Update postinstall.md for LUKS on EndeavourOS

Sandybunting:
Added a bullet point to the 'Making modules load on early boot' section to share the method I used to make the apple-bce module load early on EndeavourOS (which uses dracut, not mkinitcpio) so that the hard drive encryption password can be entered on the Apple keyboard.

### DIFF
--- a/docs/guides/postinstall.md
+++ b/docs/guides/postinstall.md
@@ -83,6 +83,16 @@ The steps to be followed vary depending upon the initramfs module loading mechan
          ```
 
     3. Run `sudo mkinitcpio -P`.
+ 
+- On EndeavourOS, which uses `dracut`
+
+    1. Create and edit a file at `/etc/dracut.conf.d/myflags.conf`, using e.g. `sudo vim /etc/dracut.conf.d/myflags.conf`
+  
+    2. Add the following line:
+
+         `force_drivers+=" apple-bce "`
+       
+    3. Run `sudo dracut --force` to regenerate the initramfs with this change
 
 - On systems with other initramfs/initrd generation systems:
 


docs/guides/postinstall.md
```diff
@@ -83,6 +83,16 @@ The steps to be followed vary depending upon the initramfs module loading mechan
          ```
 
     3. Run `sudo mkinitcpio -P`.
+ 
+- On EndeavourOS, which uses `dracut`
```
AdityaGarg8, line 87: Better to use this line:

On systems with `dracut` (Commonly used on EndeavourOS and Fedora):

docs/guides/postinstall.md
```diff
@@ -83,6 +83,16 @@ The steps to be followed vary depending upon the initramfs module loading mechan
...
+
+    1. Create and edit a file at `/etc/dracut.conf.d/myflags.conf`, using e.g. `sudo vim /etc/dracut.conf.d/myflags.conf`
```
AdityaGarg8, line 89: Create and edit a file? Isn't the file available by default?

If not available, just do `echo "force_drivers+=\" apple-bce \"" | sudo tee /etc/dracut.conf.d/myflags.conf`

docs/guides/postinstall.md
```diff
@@ -83,6 +83,16 @@ The steps to be followed vary depending upon the initramfs module loading mechan
...
+  
+    2. Add the following line:
+
+         `force_drivers+=" apple-bce "`
```
AdityaGarg8, line 93: Please use correct way to enter code:

         ```sh
         force_drivers+=" apple-bce "
         ```

AdityaGarg8 (changes requested):

docs/guides/postinstall.md
Sandybunting: No, you do have to create the file yourself for some reason.

NoaHimesaka1873: I think the file should be named "t2linux-modules.conf" for clarity. Otherwise LGTM

NoaHimesaka1873:
@Sandybunting check my comment.

AdityaGarg8:
Also fix the lint errors please.

PR #525, opened 2024-05-04
tools: improve firmware script

soopyc:
This PR contains a number of improvements to the firmware script.

1. verbosity logging is now handled with bash's `getopts` builtin to reduce repetition and to improve readability
2. situations where `/lib` doesn't exist (e.g. nixos) is handled by ~~`mkdir -p` before anything attempts to access it~~ quitting the script to avoid breaking other distros
3. The python logging level snippet was added since it was dumping a bunch of errors (repeated firmware) which I am pretty sure is harmless

**I have tested this on**
- Macbook16,2
  - [x] macOS 14.4.1
  - [x] NixOS live ISO (v6.4.9-3)
  - [x] Fedora live ISO (F40 stable v0)

### DIFF
--- a/docs/tools/firmware.sh
+++ b/docs/tools/firmware.sh
@@ -8,6 +8,16 @@
 """:"
 set -euo pipefail
 
+verbose=""
+while getopts "vhx" option; do
+	case $option in
+		v) verbose="-v" ;;
+		h) echo "usage: $0 [-vhx]"; exit 0 ;;
+		x) set -x;;
+		?) exit 1 ;; 
+	esac
+done
+
 os=$(uname -s)
 case "$os" in
 	(Darwin)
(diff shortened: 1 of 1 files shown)


soopyc:
Force-pushed to fix the superfluous `\?` escape.

Also tested working on Fedora, so I think this is good enough:tm:.

docs/tools/firmware.sh
```diff
@@ -51,46 +49,37 @@ case "$os" in
...
-			sudo tar --warning=no-unknown-keyword -xC $workdir -f $mountpoint/firmware.tar.gz
-			sudo python3 $0 $workdir $workdir/firmware-renamed.tar &> /dev/null
-			sudo tar -xC /lib/firmware -f $workdir/firmware-renamed.tar
-		fi
+		sudo mount ${verbose} /dev/nvme0n1p1 $mountpoint
+		sudo tar --warning=no-unknown-keyword ${verbose} -xC $workdir -f $mountpoint/firmware.tar.gz
+		sudo python3 $0 $workdir $workdir/firmware-renamed.tar ${verbose}
+
+		# ensure /lib/firmware exists before doing this
+		sudo mkdir ${verbose} -p /lib/firmware
```
sharpenedblade, line 64: We shouldn't do this since `/lib` (and `/usr/lib`) is usually managed by the distro, you should just make the script exit if `/lib/firmware` doesnt exist.

sharpenedblade (changes requested):

docs/tools/firmware.sh
soopyc: That does make sense but breaks setups on NixOS. Perhaps falling back to copying the firmware to some other directory then existing would be a better solution?

sharpenedblade: > That does make sense but breaks setups on NixOS.

NixOS can ship a custom script, its weird enough that our script will probably break something anyways. You could make it print a warning to see the wiki or something.

> Perhaps falling back to copying the firmware to some other directory then exiting would be a better solution?

This wont help much since the firmware archive is in the EFI partition, so if somebody wanted the firmware they can just extract it from `/boot/efi`.

soopyc: > NixOS can ship a custom script, its weird enough that our script will probably break something anyways. You could make it print a warning to see the wiki or something.

Alright, will see what can be done.

> This wont help much since the firmware archive is in the EFI partition, so if somebody wanted the firmware they can just extract it from `/boot/efi`.

AFAIK the original firmware stuff is unrenamed and (probably? - will test now) doesn't work OOTB. That was one of my motivations of making this PR.

sharpenedblade: > AFAIK the original firmware stuff is unrenamed and (probably? - will test now) doesn't work OOTB. That was one of my motivations of making this PR.

The firmware archive needs renaming, but after renaming it should *just work*. The NixOS script can just be a fork of this script with NixOS specific stuff added, bt NixOS is currently unmaintained so I dont know if this will happen soon.

sharpenedblade (approved):

PR #530, opened 2024-05-07
Fix typo in "Running script directly on Linux" warning section

s4Dt0y:


AdityaGarg8:
Doesn't look like a typo tbh, but still the change looks good!

PR #539, opened 2024-05-24
Arch: Update installation guide

AdityaGarg8:
I found some errors when trying to install Arch, so correcting them

### DIFF
--- a/docs/distributions/arch/installation.md
+++ b/docs/distributions/arch/installation.md
@@ -14,7 +14,7 @@ You will need:
 3. Follow the Arch Wiki guide from [here](https://wiki.archlinux.org/index.php/Installation_guide#Set_the_console_keyboard_layout) up to "Format the partitions".
 
     1. You will need to reformat your partitions, except for the EFI partition. The other partitions will need to reformatted as described in the Arch Wiki Installation guide. For the EFI system partition (mentioned in a note on the Arch Wiki), there will be one at `/dev/nvme0n1p1` and you can use this if you don't intend to install Windows or already have it installed. If you do intend to triple boot, refer to [this guide](https://wiki.t2linux.org/guides/windows/).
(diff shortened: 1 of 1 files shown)


AdityaGarg8:
Add apple-bce to the MODULES list in /etc/mkinitcpio.conf, and then run mkinitcpio -P

this step doesn't seem to be necessary. In fact it was causing panics for me.

Redecorating (approved):
I haven't tested this but it makes sense to me.

Redecorating:
> Add apple-bce to the MODULES list in /etc/mkinitcpio.conf, and then run mkinitcpio -P
> 
> this step doesn't seem to be necessary. In fact it was causing panics for me.

hmm this is for if you want apple-bce to be loaded before the root partition mounts (eg if its encrypted and you need to type the password). Do you have a copy of the kernel panic log?

AdityaGarg8:
> > Add apple-bce to the MODULES list in /etc/mkinitcpio.conf, and then run mkinitcpio -P
> > this step doesn't seem to be necessary. In fact it was causing panics for me.
> 
> hmm this is for if you want apple-bce to be loaded before the root partition mounts (eg if its encrypted and you need to type the password). Do you have a copy of the kernel panic log?

Nope. I've removed Arch 😞. Still, I don't find this very necessary. If you want, I can keep this

Redecorating:
> > > Add apple-bce to the MODULES list in /etc/mkinitcpio.conf, and then run mkinitcpio -P
> > > this step doesn't seem to be necessary. In fact it was causing panics for me.
> > 
> > 
> > hmm this is for if you want apple-bce to be loaded before the root partition mounts (eg if its encrypted and you need to type the password). Do you have a copy of the kernel panic log?
> 
> Nope. I've removed Arch 😞. Still, I don't find this very necessary. If you want, I can keep this

It doesn't cause issues for me so maybe we can keep it for the time being.

PR #548, opened 2024-06-12
Firmware: Improvements in getting firmware directly from macOS on Linux

AdityaGarg8:


AdityaGarg8:
@sharpenedblade @soopyc please test.

docs/tools/firmware.sh
```diff
@@ -577,49 +577,58 @@ case "$os" in
...
-				dmg2img_check
 				# Load the apfs driver, and install if missing
-				sudo modprobe apfs >/dev/null 2>&1 || apfs_install
+				sudo modprobe ${verbose} apfs 2>/dev/null || apfs_install
 				unmount_macos_and_cleanup () {
 					sudo rm -r ${verbose} ${workdir} || true
-					sudo umount ${verbose} ${loopdevice} || true
-					sudo umount ${verbose} ${macosvol} || true
-					sudo rm -r ${verbose} ${imgdir} || true
+					for i in 0 1 2 3 4 5
```
Redecorating, line 585: If the user has made extra apfs volumes could this cause issues? I think i have a couple extra volumes so I'll test it when I boot to Linux in a minute.

AdityaGarg8: I had tried making extra Volumes, they get the number from 6. Still, it would be better to test it from your side.

Redecorating: hmm seems to work fine for me

Redecorating (approved):

docs/tools/firmware.sh
AdityaGarg8: > hmm seems to work fine for me

Any error you encountered if you run it with -v

Redecorating: just errors about some of the apfs volumes not being mountable but that's expected

AdityaGarg8: Thanks.

PR #549, opened 2024-06-14
Cleanup and refactor firmware.sh

sharpenedblade:
There are minor visual changes and some of the "warnings" about installing packages got removed but most of the logic is exactly the same. `9c427e2` has a gnarly diff but the code changes are actually really small.

### DIFF
--- a/docs/tools/firmware.sh
+++ b/docs/tools/firmware.sh
@@ -5,609 +5,8 @@
 #
 # The python script is based upon the original work by The Asahi Linux Contributors.
 
-""":"
-set -euo pipefail
-
-verbose=""
-while getopts "vhx" option; do
-	case $option in
-		v) verbose="-v" ;;
-		h) echo "usage: $0 [-vhx]"; exit 0 ;;
-		x) set -x;;
-		?) exit 1 ;; 
-	esac
-done
-
-aur_install() {
-	local aur_package=$1
-	dir=$(mktemp -d)
-	cd "$dir"
-	sudo pacman -Sy --noconfirm git base-devel
-	git clone "https://aur.archlinux.org/$aur_package.git" .
-	makepkg -si --noconfirm
-	cd - >/dev/null
-	sudo rm -r ${verbose} "$dir"
-}
-
-homebrew_check () {
-	if [ ! -f "/usr/local/bin/brew" ]
-	then
-		echo -e "\nHomebrew not found!"
-		echo
-		read -rp "Press enter to install Homebrew."
-		/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
-	fi
-}
-
-detect_package_manager () {
-	if [[ $(uname -s) == "Darwin" ]]
-	then
-		echo brew
-	elif apt --help >/dev/null 2>&1
-	then
-		echo apt
-	elif dnf >/dev/null 2>&1
-	then
-		echo dnf
-	elif pacman -h >/dev/null 2>&1
-	then
-		echo pacman
-	else
-		echo NONE
-	fi
-}
-
-install_package() {
-	local package=$1
-	local package_manager
-	package_manager=$(detect_package_manager)
-	echo -e "$package is missing!\n"
-	read -p "Press enter to install $package and its dependencies. Alternatively you can terminate this script by pressing Control+C and install $package yourself, if you want to install it via some alternate method."
-
-	case $package_manager in
-		"apt")
-			case $package in
-				"linux-apfs-rw")
-					sudo apt upgrade
-					sudo apt install --reinstall -y apfs-dkms
-					sudo modprobe apfs
-					;;
-				*)
-					sudo apt upgrade
-					sudo apt install -y "$package" ;;
-			esac ;;
-		"dnf")
-			case $package in
-				"linux-apfs-rw")
-					sudo dnf -y copr enable sharpenedblade/t2linux
-					sudo dnf install -y linux-apfs-rw
-					echo -e "\nRunning akmods\n"
-					sudo akmods
-					sudo modprobe apfs
-					;;
-				*)
-					sudo dnf install -y "$package" ;;
-			esac ;;
-		"pacman")
-			case $package in
-				"linux-apfs-rw")
-					aur_install linux-apfs-rw-dkms-git
-					sudo modprobe apfs
-					;;
-				"dmg2img")
-					aur_install dmg2img ;;
-				*)
-					sudo pacman -Sy --noconfirm "$package" ;;
-			esac ;;
-		"brew")
-			case $package in
-				*)
-					homebrew_check
-					brew install "$package"
-					;;
-			esac ;;
-		"NONE")
-			read -p "The script could not detect your package manager. Please install $package manually and press enter once you have it installed." ;;
-	esac
-}
-
-create_firmware_archive() {
-	py_script=$1
-	firmware_tree=$2
-	archive=$3
-	python_check
-	python3 "$py_script" "$firmware_tree" "$archive" ${verbose}
-	if [[ $(uname -s) == "Darwin" ]]; then
-		identifier=$(system_profiler SPHardwareDataType | grep "Model Identifier" | cut -d ":" -f 2 | xargs)
-		if [[ (${identifier} = iMac19,1) || (${identifier} = iMac19,2) || (${identifier} = iMacPro1,1) ]]; then
-			nvramfile=$(ioreg -l | grep RequestedFiles | cut -d "/" -f 5 | rev | cut -c 4- | rev)
-			txcapblob=$(ioreg -l | grep RequestedFiles | cut -d "/" -f 3 | cut -d "\"" -f 1)
-			cp ${verbose} $firmware_tree/wifi/C-4364__s-B2/${nvramfile} \
-				brcmfmac4364b2-pcie.txt
-			cp ${verbose} $firmware_tree/wifi/C-4364__s-B2/${txcapblob} \
-				brcmfmac4364b2-pcie.txcap_blob
-			tar --append ${verbose} -f $archive \
-				brcmfmac4364b2-pcie.txt brcmfmac4364b2-pcie.txcap_blob
-			rm ${verbose} brcmfmac4364b2-pcie.txt brcmfmac4364b2-pcie.txcap_blob
-		fi
-	fi
-}
-reload_kernel_modules () {
-	echo "Reloading Wi-Fi and Bluetooth drivers"
-	sudo modprobe -r ${verbose} brcmfmac_wcc || true
-	sudo modprobe -r ${verbose} brcmfmac || true
-	sudo modprobe ${verbose} brcmfmac || true
-	sudo modprobe -r ${verbose} hci_bcm4377 || true
-	sudo modprobe ${verbose} hci_bcm4377 || true
-}
-
-python_check () {
-	if ! [[ $(uname -s) = "Darwin" ]]; then
-		return 0
-	fi
-	if [ ! -f "/Library/Developer/CommandLineTools/usr/bin/python3" ] && [ ! -f "/Applications/Xcode.app/Contents/Developer/usr/bin/python3" ]
-	then
-		echo -e "\nPython 3 not found. You will be prompted to install Xcode command line developer tools."
-		xcode-select --install
-		echo
-		read -p "Press enter after you have installed Xcode command line developer tools."
-	fi
-}
-
-create_deb () {
-	if [ ! -f "/usr/local/bin/dpkg" ]
-	then
-		install_package dpkg
-	fi
-
-	echo -e "\nBuilding deb package"
-	workarea=$(mktemp -d)
-	create_firmware_archive "$0" /usr/share/firmware ${workarea}/firmware.tar
-	cd ${workarea}
-	mkdir -p deb
-	cd deb
-	mkdir -p DEBIAN
-	mkdir -p usr/lib/firmware/brcm
-	cd usr/lib/firmware/brcm
-	tar -xf ${workarea}/firmware.tar ${verbose}
-	cd - >/dev/null
-
-	cat <<- EOF > DEBIAN/control
-		Package: apple-firmware
-		Version: ${ver}-1
-		Maintainer: Apple
-		Architecture: all
-		Description: Wi-Fi and Bluetooth firmware for T2 Macs
-	EOF
-
-	cat <<- EOF > DEBIAN/postinst
-		modprobe -r brcmfmac_wcc || true
-		modprobe -r brcmfmac || true
-		modprobe brcmfmac || true
-		modprobe -r hci_bcm4377 || true
-		modprobe hci_bcm4377 || true
-	EOF
-
-	chmod a+x DEBIAN/control
-	chmod a+x DEBIAN/postinst
-
-	cd ${workarea}
-	if [[ ${verbose} = -v ]]
-	then
-		dpkg-deb --build --root-owner-group -Zgzip deb
-		dpkg-name deb.deb
-	else
-		dpkg-deb --build --root-owner-group -Zgzip deb >/dev/null || echo "Failed to make deb package. Run the script with -v to get logs."
-		dpkg-name deb.deb >/dev/null
-	fi
-
-	cp ${verbose} apple-firmware_${ver}-1_all.deb $HOME/Downloads
-	echo -e "\nCleaning up"
-	rm -r ${verbose} ${workarea}
-
-	echo -e "\nDeb package apple-firmware_${ver}-1_all.deb has been saved to Downloads!"
-	echo "Copy it to Linux and install it by running the following in the Linux terminal:"
-	echo -e "\nsudo apt install /path/to/apple-firmware_${ver}-1_all.deb"
-}
-
-create_rpm () {
-	if [ ! -f "/usr/local/bin/rpmbuild" ]
-	then
-		 install_package rpm
-	fi
-
-	echo -e "\nBuilding rpm package"
-	mkdir -p $HOME/rpmbuild/{BUILD,RPMS,SOURCES,SPECS,SRPMS}
-
-	# Extract firmware 
-	create_firmware_archive "$0" /usr/share/firmware $HOME/rpmbuild/SOURCES/firmware.tar
-	cd $HOME/rpmbuild/BUILD
-
-	# Create the spec file
-	cat <<- EOF > $HOME/rpmbuild/SPECS/apple-firmware.spec
-		Name:       apple-firmware
-		Version:    ${ver}
-		Release:    1
-		Summary:    Wi-Fi and Bluetooth firmware for T2 Macs
-		License:    Proprietary
-		BuildArch:  noarch
-
-		Source1: firmware.tar
-
-		%description
-		Wi-Fi and Bluetooth firmware for T2 Macs
-
-		%prep
-		tar -xf %{SOURCE1}
-
-		%build
-
-		%install
-		mkdir -p %{buildroot}/usr/lib/firmware/brcm
-		install -m 644 * %{buildroot}/usr/lib/firmware/brcm
-
-		%posttrans
-		modprobe -r brcmfmac_wcc || true
-		modprobe -r brcmfmac || true
-		modprobe brcmfmac || true
-		modprobe -r hci_bcm4377 || true
-		modprobe hci_bcm4377 || true
-
-		%files
-		/usr/lib/firmware/brcm/*
-	EOF
-
-	# Build
-	if [[ ${verbose} = -v ]]
-	then
-		rpmbuild -bb --define '_target_os linux' $HOME/rpmbuild/SPECS/apple-firmware.spec
-	else
-		rpmbuild -bb --define '_target_os linux' $HOME/rpmbuild/SPECS/apple-firmware.spec >/dev/null 2>&1 || echo "Failed to make rpm package. Run the script with -v to get logs."
-	fi
-
-	# Copy and Cleanup
-	cp ${verbose} $HOME/rpmbuild/RPMS/noarch/apple-firmware-${ver}-1.noarch.rpm $HOME/Downloads
-	echo -e "\nCleaning up"
-	rm -r ${verbose} $HOME/rpmbuild
-
-	echo -e "\nRpm package apple-firmware-${ver}-1.noarch.rpm has been saved to Downloads!"
-	echo "Copy it to Linux and install it by running the following in the Linux terminal:"
-	echo -e "\nsudo dnf install --disablerepo=* /path/to/apple-firmware-${ver}-1.noarch.rpm"
-}
-
-create_arch_pkg () {
-	if [ ! -f "/usr/local/bin/makepkg" ]
-	then
-		install_package makepkg
-	fi
-	if [ ! -f "/usr/local/bin/sha256sum" ]
-	then
-		install_package coreutils
-	fi
-
-	echo -e "\nBuilding pacman package"
-	workarea=$(mktemp -d)
-	create_firmware_archive "$0" /usr/share/firmware ${workarea}/firmware.tar
-	cd ${workarea}
-
-	# Create the PKGBUILD
-	cat <<- EOF > PKGBUILD
-		pkgname=apple-firmware
-		pkgver=${ver}
-		pkgrel=1
-		pkgdesc="Wi-Fi and Bluetooth Firmware for T2 Macs"
-		arch=("any")
-		url=""
-		license=('unknown')
-		replaces=('apple-bcm-wifi-firmware')
-		source=("firmware.tar")
-		noextract=("firmware.tar")
-		sha256sums=('SKIP')
-	EOF
-	cat <<- 'EOF' >> PKGBUILD
-
-		package() {
-			mkdir -p $pkgdir/usr/lib/firmware/brcm
-			cd $pkgdir/usr/lib/firmware/brcm
-			tar xf $srcdir/firmware.tar
-		}
-
-		install=apple-firmware.install
-	EOF
-
-	cat <<- EOF > apple-firmware.install
-		post_install() {
-			modprobe -r brcmfmac_wcc || true
-			modprobe -r brcmfmac || true
-			modprobe brcmfmac || true
-			modprobe -r hci_bcm4377 || true
-			modprobe hci_bcm4377 || true
-		}
-	EOF
-
-	# Set path to use newer bsdtar and GNU touch
-	PATH_OLD=$PATH
-	PATH=/usr/local/Cellar/libarchive/$(ls /usr/local/Cellar/libarchive | head -n 1)/bin:/usr/local/opt/coreutils/libexec/gnubin:$PATH_OLD
-
-	# Build
-	if [[ ${verbose} = -v ]]
-	then
-		PKGEXT='.pkg.tar.zst' makepkg
-	else
-		PKGEXT='.pkg.tar.zst' makepkg >/dev/null 2>&1 || echo "Failed to make pacman package. Run the script with -v to get logs."
-	fi
-
-	# Revert path to its original form
-	PATH=${PATH_OLD}
-
-	# Copy to Downloads and cleanup
-	cp ${verbose} apple-firmware-${ver}-1-any.pkg.tar.zst $HOME/Downloads
-	echo -e "\nCleaning up"
-	rm -r ${verbose} ${workarea}
-
-	echo -e "\nPacman package apple-firmware-${ver}-1-any.pkg.tar.zst has been saved to Downloads!"
-	echo "Copy it to Linux and install it by running the following in the Linux terminal:"
-	echo -e "\nsudo pacman -U /path/to/apple-firmware-${ver}-1-any.pkg.tar.zst"
-}
-
-os=$(uname -s)
-case "$os" in
-	(Darwin)
-		echo "Detected macOS"
-		ver=$(sw_vers -productVersion)
-		ver_check=$(sw_vers -productVersion | cut -d "." -f 1)
-		identifier=$(system_profiler SPHardwareDataType | grep "Model Identifier" | cut -d ":" -f 2 | xargs)
-		if [[ ${ver_check} < 12 ]] && [[ (${identifier} = MacBookPro15,4) || (${identifier} = MacBookPro16,3) || (${identifier} = MacBookAir9,1) ]]
-		then
-			cat <<- EOF
-
-			Warning: You are running a macOS version earlier than macOS Monterey.
-			Your Mac model needs Bluetooth firmware for in addition to Wi-Fi firmware, which is available only on macOS Monterey or later.
-			Only Wi-Fi firmware shall be copied to Linux.
-			For Bluetooth firmware, you can either:
-
-			a) Upgrade macOS to Monterey or later.
-			b) Run this script directly in Linux and choose the option to Download a macOS Recovery Image from there.
-
-			EOF
-		fi
-		echo -e "\nHow do you want to copy the firmware to Linux?"
-		echo -e "\n1. Copy the firmware to the EFI partition and run the same script on Linux to retrieve it."
-		echo "2. Create a tarball of the firmware and extract it to Linux."
-		echo "3. Create a Linux specific package which can be installed using a package manager."
-		echo -e "\nNote: Option 2 and 3 require additional software like python3 and tools specific for your package manager. Requirements will be told as you proceed further."
-		read choice
-		case ${choice} in
-			(1)
-				echo -e "\nMounting the EFI partition"
-				EFILABEL=$(diskutil info disk0s1 | grep "Volume Name" | cut -d ":" -f 2 | xargs)
-				sudo diskutil mount disk0s1
-				echo "Getting Wi-Fi and Bluetooth firmware"
-				tar ${verbose} -cf "/Volumes/${EFILABEL}/firmware.tar" -C /usr/share/firmware/ .
-				gzip ${verbose} --best "/Volumes/${EFILABEL}/firmware.tar"
-				if [[ (${identifier} = iMac19,1) || (${identifier} = iMac19,2) || (${identifier} = iMacPro1,1) ]]
-		then
-					nvramfile=$(ioreg -l | grep RequestedFiles | cut -d "/" -f 5 | rev | cut -c 4- | rev)
-					txcapblob=$(ioreg -l | grep RequestedFiles | cut -d "/" -f 3 | cut -d "\"" -f 1)
-					cp ${verbose} /usr/share/firmware/wifi/C-4364__s-B2/${nvramfile} "/Volumes/${EFILABEL}/brcmfmac4364b2-pcie.txt"
-					cp ${verbose} /usr/share/firmware/wifi/C-4364__s-B2/${txcapblob} "/Volumes/${EFILABEL}/brcmfmac4364b2-pcie.txcap_blob"
-				fi
-				echo "Copying this script to EFI"
-				cp "$0" "/Volumes/${EFILABEL}/firmware.sh" 2>/dev/null || curl -s https://wiki.t2linux.org/tools/firmware.sh > "/Volumes/${EFILABEL}/firmware.sh" || (echo -e "\nFailed to copy script.\nPlease copy the script manually to the EFI partition using Finder\nMake sure the name of the script is firmware.sh in the EFI partition\n" && echo && read -p "Press enter after you have copied" && echo)
-				echo "Unmounting the EFI partition"
-				sudo diskutil unmount "/Volumes/${EFILABEL}/"
-				echo
-				echo -e "Run the following commands or run this script itself in Linux now to set up Wi-Fi :-\n\nsudo mkdir -p /tmp/apple-wifi-efi\nsudo mount /dev/nvme0n1p1 /tmp/apple-wifi-efi\nbash /tmp/apple-wifi-efi/firmware.sh\nsudo umount /tmp/apple-wifi-efi\n"
-				;;
-			(2)
-
-				echo -e "\nCreating a tarball of the firmware"
-				create_firmware_archive "$0" /usr/share/firmware $HOME/Downloads/firmware.tar ${verbose}
-				echo -e "\nFirmware tarball saved to Downloads!"
-				echo -e "\nExtract the tarball contents to /lib/firmware/brcm in Linux and run the following in the Linux terminal:"
-				echo -e "\nsudo modprobe -r brcmfmac_wcc"
-				echo "sudo modprobe -r brcmfmac"
-				echo "sudo modprobe brcmfmac"
-				echo "sudo modprobe -r hci_bcm4377"
-				echo "sudo modprobe hci_bcm4377"
-				;;
-			(3)
-				echo -e "\nWhat package manager does your Linux distribution use?"
-				echo -e "\n1. apt"
-				echo "2. dnf"
-				echo "3. pacman"
-				read package
-				case ${package} in
-					(1)
-						create_deb
-						;;
-					(2)
-						create_rpm
-						;;
-					(3)
-						create_arch_pkg
-						;;
-					(*)
-						echo -e "\nError: Invalid option!"
-						exit 1
-						;;
-					esac
-				;;
-			(*)
-				echo -e "\nError: Invalid option!"
-				exit 1
-				;;
-			esac
-		;;
-
-	(Linux)
-		echo "Detected Linux"
-		if [[ ! -e /lib/firmware/brcm ]]; then
-			echo "/lib/firmware/brcm does not seem to exist. This script requires that directory to function."
-			echo "If you are on some exotic distro like NixOS, please check the wiki for more information:"
-			echo "  https://wiki.t2linux.org"
-			echo "Exiting..."
-			exit 1
-		fi
-		echo -e "\nHow do you want to copy the firmware to Linux?"
-		echo -e "\n1. Retrieve the firmware from the EFI partition."
-		echo "2. Retrieve the firmware directly from macOS."
-		echo "3. Download a macOS Recovery Image from Apple and extract the firmware from there."
-		echo -e "\nNote: If you are choosing Option 1, then make sure you have run the same script on macOS before and chose Option 1 (Copy the firmware to the EFI partition and run the same script on Linux to retrieve it) there."
-		read choice
-		case ${choice} in
-			(1)
-				echo -e "\nRe-mounting the EFI partition"
-				mountpoint=$(mktemp -d)
-				workdir=$(mktemp -d)
-				echo "Installing Wi-Fi and Bluetooth firmware"
-				sudo mount ${verbose} /dev/nvme0n1p1 $mountpoint
-				sudo tar --warning=no-unknown-keyword ${verbose} -xC ${workdir} -f $mountpoint/firmware-raw.tar.gz
-				sudo chown -R $USER ${workdir}
-				create_firmware_archive "$0" ${workdir} ${workdir}/firmware-renamed.tar ${verbose}
-
-				sudo tar ${verbose} -xC /lib/firmware/brcm -f ${workdir}/firmware-renamed.tar
-
-				for file in "$mountpoint/brcmfmac4364b2-pcie.txt" \
-					    "$mountpoint/brcmfmac4364b2-pcie.txcap_blob"
-				do
-					if [ -f "$file" ]
-					then
-						sudo cp ${verbose} $file /lib/firmware/brcm
-					fi
-				done
-				reload_kernel_modules
-				echo -e "\nKeeping a copy of the firmware and the script in the EFI partition shall allow you to set up Wi-Fi again in the future by running this script or the commands told in the macOS step in Linux only, without the macOS step."
-				read -p "Do you want to keep a copy? (y/N)" input
-				if [[ ($input != y) && ($input != Y) ]]
-				then
-					echo -e "\nRemoving the copy from the EFI partition"
-					for file in "$mountpoint/brcmfmac4364b2-pcie.txt" \
-					            "$mountpoint/brcmfmac4364b2-pcie.txcap_blob" \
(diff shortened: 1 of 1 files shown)


docs/tools/firmware.sh
```diff
@@ -5,414 +5,687 @@
...
-			echo -e "\nInstalling curl and its dependencies"
-			case ${PACKAGE_MANAGER} in
-				(apt)
-					sudo apt update
-					sudo apt install -y curl
+install_package() {
+	local package=$1
+	local package_manager
+	package_manager=$(detect_package_manager)
+	read -rp "Press enter to install $package and its dependencies. Alternatively you can terminate this script by pressing Control+C and install $package yourself, if you want to install it via some alternate method."
```
AdityaGarg8, line 391: Where is the statement notifying that the package is missing?

docs/tools/firmware.sh
```diff
@@ -5,414 +5,687 @@
...
+		echo -e "\nHomebrew not found!"
+		echo
+		read -rp "Press enter to install Homebrew."
+		/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+	fi
 }
 
 detect_package_manager () {
-	if $(apt --help >/dev/null 2>&1)
+	if [[ $(uname) == "Darwin" ]]
```
AdityaGarg8, line 370: Use uname -s

docs/tools/firmware.sh
```diff
@@ -5,414 +5,687 @@
...
+		"NONE")
+			read -rp "The script could not detect your package manager. Please install $package manually and press enter once you have it installed." ;;
+	esac
+}
+
+create_firmware_archive() {
+	firmware_tree=$1
+	archive=$2
+	rename_firmware "$firmware_tree" "$archive" ${verbose}
+	if [[ $(uname) == "Darwin" ]]; then
```
AdityaGarg8, line 443: Uname -s

docs/tools/firmware.sh
```diff
@@ -5,414 +5,687 @@
...
+	
+	sudo pacman -U /path/to/apple-firmware-${ver}-1-any.pkg.tar.zst
+	EOF
+}
+
+os=$(uname -s)
+case "$os" in
+	(Darwin)
+		if [[ $verbose = "-v" ]]; then
+			echo "Detected macOS"
```
AdityaGarg8, line 681: This should be shown even if verbose isn't on

docs/tools/firmware.sh
```diff
@@ -515,103 +786,97 @@ case "$os" in
 		;;
 
 	(Linux)
-		echo "Detected Linux"
+		if [[ $verbose = "-v" ]]; then
+			echo "Detected Linux"
```
AdityaGarg8, line 790: Again, show this even if verbose is disabled

docs/tools/firmware.sh
```diff
@@ -515,103 +786,97 @@ case "$os" in
...
 			exit 1
 		fi
-		echo -e "\nHow do you want to copy the firmware to Linux?"
-		echo -e "\n1. Retrieve the firmware from the EFI partition."
-		echo "2. Retrieve the firmware directly from macOS."
-		echo "3. Download a macOS Recovery Image from Apple and extract the firmware from there."
-		echo -e "\nNote: If you are choosing Option 1, then make sure you have run the same script on macOS before and chose Option 1 (Copy the firmware to the EFI partition and run the same script on Linux to retrieve it) there."
-		read choice
+		cat <<- EOF
+		How do you want to copy the firmware to Linux?
```
AdityaGarg8, line 802: A blank line before this information should be good

docs/tools/firmware.sh
```diff
@@ -515,103 +786,97 @@ case "$os" in
...
+		2. Retrieve the firmware directly from macOS.
+		3. Download a macOS Recovery Image from Apple and extract the firmware from there.
+		
+		Note: If you are choosing Option 1, then make sure you have run the same script on macOS before and chose Option 1 (Copy the firmware to the EFI partition and run the same script on Linux to retrieve it) there.
+		EOF
+		read -r choice
 		case ${choice} in
 			(1)
-				echo -e "\nRe-mounting the EFI partition"
+				echo -e "Re-mounting the EFI partition"
```
AdityaGarg8, line 813: Why has a blank line been removed?

docs/tools/firmware.sh
```diff
@@ -515,103 +786,97 @@ case "$os" in
...
-				echo "Reloading Wi-Fi and Bluetooth drivers"
-				sudo modprobe -r ${verbose} brcmfmac_wcc || true
-				sudo modprobe -r ${verbose} brcmfmac || true
-				sudo modprobe ${verbose} brcmfmac || true
-				sudo modprobe -r ${verbose} hci_bcm4377 || true
-				sudo modprobe ${verbose} hci_bcm4377 || true
-				echo -e "\nKeeping a copy of the firmware and the script in the EFI partition shall allow you to set up Wi-Fi again in the future by running this script or the commands told in the macOS step in Linux only, without the macOS step."
-				read -p "Do you want to keep a copy? (y/N)" input
+				reload_kernel_modules
+				echo "Keeping a copy of the firmware and the script in the EFI partition shall allow you to set up Wi-Fi again in the future by running this script or the commands told in the macOS step in Linux only, without the macOS step."
```
AdityaGarg8, line 824: Again, why removed the blank line

docs/tools/firmware.sh
```diff
@@ -515,103 +786,97 @@ case "$os" in
...
+				read -rp "Do you want to keep a copy? (y/N)" input
 				if [[ ($input != y) && ($input != Y) ]]
 				then
-					echo -e "\nRemoving the copy from the EFI partition"
+					echo -e "Removing the copy from the EFI partition"
```
AdityaGarg8, line 828: Again blank line removed

docs/tools/firmware.sh
```diff
@@ -515,103 +786,97 @@ case "$os" in
...
 					done
 				fi
-				sudo rm -r ${verbose} ${workdir}
-				sudo umount $mountpoint
-				sudo rmdir $mountpoint
-				echo -e "\nDone!"
+				sudo rm -r ${verbose} "${workdir}"
+				sudo umount "$mountpoint"
+				sudo rmdir "$mountpoint"
+				echo -e "Done!"
```
AdityaGarg8, line 843: Blank line removed

docs/tools/firmware.sh
```diff
@@ -515,103 +786,97 @@ case "$os" in
...
-							sudo umount ${macosdir}/vol${i} 2>/dev/null || true
+							sudo umount "${macosdir}/vol${i}" 2>/dev/null || true
 						fi
 					done
-					sudo rm -r ${verbose} ${macosdir} || true
+					sudo rm -r ${verbose} "${macosdir}" || true
 				}
 
-				echo -e "\nMounting the macOS volume"
+				echo -e "Mounting the macOS volume"
```
AdityaGarg8, line 862: Blank line removed

docs/tools/firmware.sh
```diff
@@ -621,64 +886,53 @@ case "$os" in
...
-					sudo rm -r ${verbose} ${imgdir}
-					sudo losetup -d /dev/${loopdev}
+					sudo rm -r ${verbose} "${workdir}"
+					sudo umount ${verbose} "${loopdevice}"
+					sudo rm -r ${verbose} "${imgdir}"
+					sudo losetup -d /dev/"${loopdev}"
 				}
 
-				echo -e "\nDownloading macOS Recovery Image"
+				echo -e "Downloading macOS Recovery Image"
```
AdityaGarg8, line 907: Blank line removed

docs/tools/firmware.sh
```diff
@@ -621,64 +886,53 @@ case "$os" in
...
 				then
 					curl -O https://raw.githubusercontent.com/kholia/OSX-KVM/master/fetch-macOS-v2.py
 				else
 					curl -s -O https://raw.githubusercontent.com/kholia/OSX-KVM/master/fetch-macOS-v2.py
 				fi
 				echo -e "\nNote: In order to get complete firmware files, download macOS Monterey or later.\n"
 				python3 fetch-macOS-v2.py
-				echo -e "\nConverting image from .dmg to .img"
-				if [[ ${verbose} = -v ]]
+				echo -e "Converting image from .dmg to .img"
```
AdityaGarg8, line 919: Blank line removed

AdityaGarg8 (changes requested):

docs/tools/firmware.sh
AdityaGarg8: I'll prefer keeping that warning

docs/tools/firmware.sh
sharpenedblade: I wasnt sure if the extra newlines were intentional, I think all of the newline changes are reverted now.

docs/tools/firmware.sh
```diff
@@ -388,7 +388,7 @@ install_package() {
 	local package=$1
 	local package_manager
 	package_manager=$(detect_package_manager)
-	read -rp "Press enter to install $package and its dependencies. Alternatively you can terminate this script by pressing Control+C and install $package yourself, if you want to install it via some alternate method."
+	read -rp "$package is missing. Press enter to install $package and its dependencies. Alternatively you can terminate this script by pressing Control+C and install $package yourself, if you want to install it via some alternate method."
```
AdityaGarg8, line 391: I actually liked the old wording.

```
Dpkg and/or its dependencies are missing!

Press enter...
```

AdityaGarg8 (changes requested):

docs/tools/firmware.sh
sharpenedblade: The package's dependencies aren't actually missing so I didnt want to confuse users. If they see that they might think that they have dpkg installed but are missing its dependencies.

AdityaGarg8:
Previously it was possible to manually run the python script using python3 firmware.sh. It can be useful in various cases, like having some firmware archive and want to rename it. Iirc, that's why I chose a polyglot script. Will it still be possible?

AdityaGarg8:
Also you have removed many intended blank lines from the code. It becomes a pain to read what is written on the terminal without blank lines. Lots of blank lines missing in create_deb, create_rpm and create_arch_package.

AdityaGarg8:
Also, the commit history is a pain to review here. Probably commit like

Refactor Install packages.
Refactor loading and unloading of drivers.


Etc.

Squash the commits which are addressing minor mistakes made by you.

Lastly, do the miscellaneous changes like shell check fixes, refactoring the python bit etc.

AdityaGarg8:
If you want, you can open a new PR with a proper commit history. Probably shall also fix the intended blank lines, since hunting them shall a difficult.

sharpenedblade:
> Previously it was possible to manually run the python script using python3 firmware.sh. It can be useful in various cases, like having some firmware archive and want to rename it. Iirc, that's why I chose a polyglot script. Will it still be possible?

I am going to make the script accept flags to pick a method so I can just add a flag to run the rename function

> Also you have removed many intended blank lines from the code. It becomes a pain to read what is written on the terminal without blank lines. Lots of blank lines missing in create_deb, create_rpm and create_arch_package.

All of those were changed in [d0c05f8](https://github.com/t2linux/wiki/pull/549/commits/d0c05f87e63571c7a708e9571628383916dc0a6e), I can just revert that commit.

> Also, the commit history is a pain to review here. Probably commit like

I can squash down some of the commits but its not going to be that much better. Github is messing up the diff, its a lot cleaner if you run `git diff master --diff-algorithm minimal`.

docs/tools/firmware.sh
sharpenedblade: We tell the user that we are going to install its dependencies in the next sentence. That sentence reads to me like its possible that you might have dpkg but not one of its dependencies, which is not what the script is actually checking

AdityaGarg8: > We tell the user that we are going to install its dependencies in the next sentence. That sentence reads to me like its possible that you might have dpkg but not one of its dependencies, which is not what the script is actually checking

```
Dpkg is missing!

Press enter...
```

Then do this

AdityaGarg8:
Last 2 commits are absolutely unnecessary, The only thing that is good is replacing echo with cat and renaming firmware.tar to firmware-raw.tar. And please don't remove logging. It's required for have an idea where the script fails, in case it does. Also, why python_check is there in create_firmware even if it is macOS only check?

AdityaGarg8:
```
				for file in "$mountpoint/brcmfmac4364b2-pcie.txt" \
					    "$mountpoint/brcmfmac4364b2-pcie.txcap_blob"
				do
					if [ -f "$file" ]
					then
						sudo cp ${verbose} "$file" /lib/firmware/brcm
					fi
				done
```

And why has this been removed?

AdityaGarg8:
Also replace == with = in comparisons. IIRC == is bash specific, not POSIX compliant

sharpenedblade:
This isnt the latest version yet, I havent pushed my changes. I will request a review when im done.

sharpenedblade:
> And why has this been removed?

Its appended to the tar archive so its not placed in the ESP anymore.

AdityaGarg8:
> > And why has this been removed?
> 
> Its appended to the tar archive so its not placed in the ESP anymore.

It's still placed in esp, and should be placed on esp

https://github.com/t2linux/wiki/blob/d2bc82844e3da299dfad2eb1dcb56a0a6abd9a6b/docs/tools/firmware.sh#L720

AdityaGarg8:
> This isnt the latest version yet, I havent pushed my changes. I will request a review when im done.

Mark it as a draft then

sharpenedblade:
> > > And why has this been removed?
> > 
> > 
> > Its appended to the tar archive so its not placed in the ESP anymore.
> 
> It's still placed in esp

The current pushed branch is in a half-fixed state, its in the ESP on my local branch

> and should be placed on esp
Why? Its just like the other firmware files, we just need to rename it on macos

AdityaGarg8:
> > > > And why has this been removed?
> > > 
> > > 
> > > Its appended to the tar archive so its not placed in the ESP anymore.
> > 
> > 
> > It's still placed in esp
> 
> The current pushed branch is in a half-fixed state, its in the ESP on my local branch
> 
> > and should be placed on esp
> 
> Why? Its just like the other firmware files, we just need to rename it on macos

Because ESP files are renamed on Linux, not macOS

sharpenedblade:
Im not sure exactly what to put in the help output, its a bit unclear right now.

sharpenedblade:
> Because ESP files are renamed on Linux, not macOS

I put it in the firmware-raw.tar.gz bundle but copy it on linux from the extracted tarball instead of giving it to the python script.

AdityaGarg8:
> Im not sure exactly what to put in the help output, its a bit unclear right now. Im open to renaming the subcommands too.

I'd rather say add the rename only and get from macOS functionality. It would be better to give options to users rather than doing automatation. I don't think there will be a big gain by just reducing the step of choosing the method by the user. In fact passing these long parameters is a bigger pain.

AdityaGarg8:
Get from macOS should be good if you want to add some sort of systemd service which you planned before. Rename only should be good as script is no longer polyglot

AdityaGarg8:
Try to make things simple.

AdityaGarg8:
> > Because ESP files are renamed on Linux, not macOS
> 
> I put it in the firmware-raw.tar.gz bundle but copy it on linux from the extracted tarball instead of giving it to the python script.

Change was absolutely unnecessary, but it's fine if it works.

sharpenedblade:
> I don't think there will be a big gain by just reducing the step of choosing the method by the user.
> In fact passing these long parameters is a bigger pain.

There is no UX change if you run the script without any subcommands. If you dont pass anything then it uses the old code which interactively prompts the user.

> I'd rather say add the rename only and get from macOS functionality.

rename_only is the same as running the old script with `python3`. get_from_macos is just method 4 from before.

> Change was absolutely unnecessary, but it's fine if it works.

The idea was to add just firmware.tar.gz to the ESP in the future with renamed files in it.

sharpenedblade:
> Try to make things simple.

In the script in general or just the argument parsing code? Argument parsing in bash is a PITA, especially without gnu-isms. Im not sure how much simpler it can be.

AdityaGarg8:
> > I don't think there will be a big gain by just reducing the step of choosing the method by the user.
> > In fact passing these long parameters is a bigger pain.
> 
> There is no UX change if you run the script without any subcommands. If you dont pass anything then it uses the old code which interactively prompts the user.

there is no UX change even after refactoring and cleanup, still we are agreeing to it, cause it makes code less bulky. Similarly, we don't really need these automation features cause it's not useful at all.
> 
> > I'd rather say add the rename only and get from macOS functionality.
> 
> rename_only is the same as running the old script with `python3`. get_from_macos is just method 4 from before.
> 
> > Change was absolutely unnecessary, but it's fine if it works.
> 
> The idea was to add just firmware.tar.gz to the ESP in the future with renamed files in it.

AdityaGarg8:
> > Try to make things simple.
> 
> In the script in general or just the argument parsing code? Argument parsing in bash is a PITA, especially without gnu-isms. Im not sure how much simpler it can be.

What I meant by try to make things simple was to not add unnecessary features like automation

sharpenedblade:
> Similarly, we don't really need these automation features cause it's not useful at all.

We can stick it inside of a systemd service and handle everything firmware related automatically. I also want to make a (very) simple gui to download the isos and get the firmware, so I want to be able to automate the firmware script.  Its only ~20 LOC and the prompting should have been seperated from the actual code anyways.

AdityaGarg8:
The only use case for automation I feel is that first boot systemd service of yours. That too I'll suggest first testing. I guess we can do automation in a different PR.

sharpenedblade:
> What I meant by try to make things simple was to not add unnecessary features like automation

I want to eventually make the installation seamless where you can just download a gui app, click your distro, reboot into linux, follow the installer, then have a working system.

AdityaGarg8:
> > Similarly, we don't really need these automation features cause it's not useful at all.
> 
> We can stick it inside of a systemd service and handle everything firmware related automatically. I also want to make a (very) simple gui to download the isos and get the firmware, so I want to be able to automate the firmware script. Its only ~20 LOC and the prompting should have been seperated from the actual code anyways.

GUI seems to be a nice idea. But again, we can keep it for a different PR. Also, we need gui for both macOS and Linux then.

sharpenedblade:
> GUI seems to be a nice idea. But again, we can keep it for a different PR. Also, we need gui for both macOS and Linux then.

Im not going to PR that here, I just wanted to get the firmware script ready since I was already moving stuff around.

> Also, we need gui for both macOS and Linux then.

It would be in something static like rust or go and in a cross platform toolkit

AdityaGarg8:
Also, you probably can use one letter arguments if it's specifically to be run at backend as a systemd service or gui

AdityaGarg8:
Then you can do an it before read to check for the variable set by the argument and if it's missing, use read to get the variable

sharpenedblade:
> Also, you probably can use one letter arguments if it's specifically to be run at backend as a systemd service or gui. Much easy to add here.

Its 7 LOC to handle the long arguments so its not going to simplify things that much. I can change it but I thought its clearer to make them "subcommands"

AdityaGarg8:
> > Also, you probably can use one letter arguments if it's specifically to be run at backend as a systemd service or gui. Much easy to add here.
> 
> Its 7 LOC to handle the long arguments so its not going to simplify things that much. I can change it but I thought its clearer to make them "subcommands"

So as I said before, add them when you really need them.

AdityaGarg8:
I.e., when you have actually started working on your gui/systemd service

AdityaGarg8:
You probably can keep them in your local copy, but I don't find any use of them for now in the wiki,

sharpenedblade:
Il take the automation commit out and put it in a new PR with the systemd stuff

AdityaGarg8:
Wasn't python working without sudo before?

sharpenedblade:
> Wasn't python working without sudo before?

No, when extracting the EFI tarball on linux we use sudo for every other command, which messes up permissions. I tested it and it was broken, but adding sudo to python3 fixes it.

AdityaGarg8:
> > Wasn't python working without sudo before?
> 
> No, when extracting the EFI tarball on linux we use sudo for every other command, which messes up permissions. I tested it and it was broken, but adding sudo to python3 fixes it.

I don't think it's required for a method other than EFI

AdityaGarg8:
Atleast that's what's done in the current version of the script

sharpenedblade:
> Atleast that's what's done in the current version of the script

There is no good way to execute a bash function with `sudo`.

AdityaGarg8:
> > Atleast that's what's done in the current version of the script
> 
> There is no good way to execute a bash function with `sudo`.

Is there any good reason why we have switched to making the python bit a function instead of previous polyglot script.

AdityaGarg8:
$USER will still not be root if running with sudo iirc

sharpenedblade:
> > > Atleast that's what's done in the current version of the script
> > 
> > 
> > There is no good way to execute a bash function with `sudo`.

I found a not-hacky fix for the permissions thats better anyways.

> Is there any good reason why we have switched to making the python bit a function instead of previous polyglot script.

Not really, but the previous method was hacky and made us rely on $0.

sharpenedblade:
> $USER will still not be root if running with sudo iirc

That is the point, we want $workdir to be the current (non-root) user

AdityaGarg8:
> > $USER will still not be root if running with sudo iirc
> 
> That is the point, we want $workdir to be the current (non-root) user

Looks good then.

AdityaGarg8:
Also for the Fix Message commit, I dunno is $package is missing would be right for makepkg, cause the logic checks for makepkg as well as coreutils.

sharpenedblade:
> Also for the Fix Message commit, I dunno if "$package is missing" would be right for makepkg, cause the logic checks for makepkg as well as coreutils.

I split that `if` up to only install the right packages.

AdityaGarg8:
> > Also for the Fix Message commit, I dunno if "$package is missing" would be right for makepkg, cause the logic checks for makepkg as well as coreutils.
> 
> I split that `if` up to only install the right packages.

It was intentionally not split so that latest version of both is installed. Brew for some reason hasn't included coreutils as a dependency but without it makepkg doesn't work.

sharpenedblade:
> > > Also for the Fix Message commit, I dunno if "$package is missing" would be right for makepkg, cause the logic checks for makepkg as well as coreutils.
> > 
> > 
> > I split that `if` up to only install the right packages.
> 
> It was intentionally not split so that latest version of both is installed. Brew for some reason hasn't included coreutils as a dependency but without it makepkg doesn't work.

GNU coreutils are stable so it shouldnt really matter what version they are as long as it is somewhat modern.

AdityaGarg8:
> > > > Also for the Fix Message commit, I dunno if "$package is missing" would be right for makepkg, cause the logic checks for makepkg as well as coreutils.
> > > 
> > > 
> > > I split that `if` up to only install the right packages.
> > 
> > 
> > It was intentionally not split so that latest version of both is installed. Brew for some reason hasn't included coreutils as a dependency but without it makepkg doesn't work.
> 
> GNU coreutils are stable so it shouldnt really matter what version they are as long as it is somewhat modern.

Hmm but I'll prefer to not take any chances here.

AdityaGarg8:
and/or its dependencies line kinda fits here.

sharpenedblade:
> Hmm but I'll prefer to not take any chances here.

The old code did not install the latest verrsion anyways

AdityaGarg8:
> > Hmm but I'll prefer to not take any chances here.
> 
> The old code did not install the latest verrsion anyways

How can you say that?

sharpenedblade:
It will only install the latest version if makepkg or coreutils are missing, if they are both installed but out of date nothing happens.

AdityaGarg8:
> It will only install the latest version if makepkg or coreutils are missing, if they are both installed but out of date nothing happens.

If someone has both of them installed, I believe homebrew will update both together when it upgrades its packages.

AdityaGarg8:
Anyways it's a small issue, so we can deal with it later.

AdityaGarg8:
I doubt it would actually affect people

AdityaGarg8:
Commit 59e0706ca11800b545fca523b23bc4f04f630cb1 doesn't seem to be necessary since the code is not duplicated. Making function makes things unecessary. And why has loopdevice been removed?

sharpenedblade:
> Commit [59e0706](https://github.com/t2linux/wiki/commit/59e0706ca11800b545fca523b23bc4f04f630cb1) doesn't seem to be necessary since the code is not duplicated. And why has loopdevice been removed?

$loopdev and $loopdevice are exactly the same

sharpenedblade:
> Commit [59e0706](https://github.com/t2linux/wiki/commit/59e0706ca11800b545fca523b23bc4f04f630cb1) doesn't seem to be necessary since the code is not duplicated.

Its for logging to a file later

AdityaGarg8:
> > And why has loopdevice been removed?
> 
> $loopdev and $loopdevice are exactly the same

Loopdev is loop50
Loopdevice is loop50p1 or loop50p2

sharpenedblade:
> > > And why has loopdevice been removed?
> > 
> > 
> > $loopdev and $loopdevice are exactly the same
> 
> Loopdev is loop50 Loopdevice is loop50p1 or loop50p2

I didnt see that when testing, il revert the change. I think its because I used a partitionless file to check the code

AdityaGarg8:
> > > > And why has loopdevice been removed?
> > > 
> > > 
> > > $loopdev and $loopdevice are exactly the same
> > 
> > 
> > Loopdev is loop50 Loopdevice is loop50p1 or loop50p2
> 
> I didnt see that when testing, il revert the change. I think its because I used a partitionless file to check the code

Be assured there is no variable unnecessarily declared in the script.

sharpenedblade:
> > > > > And why has loopdevice been removed?
> > > > 
> > > > 
> > > > $loopdev and $loopdevice are exactly the same
> > > 
> > > 
> > > Loopdev is loop50 Loopdevice is loop50p1 or loop50p2
> > 
> > 
> > I didnt see that when testing, il revert the change. I think its because I used a partitionless file to check the code
> 
> Be assured there is no variable unnecessarily declared in the script.

I thought it was just old code since they were similarily names

AdityaGarg8:
Please squash the commit when you revert loopdev

AdityaGarg8:
> > > > > > And why has loopdevice been removed?
> > > > > 
> > > > > 
> > > > > $loopdev and $loopdevice are exactly the same
> > > > 
> > > > 
> > > > Loopdev is loop50 Loopdevice is loop50p1 or loop50p2
> > > 
> > > 
> > > I didnt see that when testing, il revert the change. I think its because I used a partitionless file to check the code
> > 
> > 
> > Be assured there is no variable unnecessarily declared in the script.
> 
> I thought it was just old code since they were similarily names

You can change the names to avoid confusion. I couldn't get a good name while writing it.

sharpenedblade:
> Please squash the commit when you revert loopdev
I squashed everything down

AdityaGarg8:
I would avoid changes to python code. If you want to get it merged, get a review from Redecorating

sharpenedblade:
> I would avoid changes to python code. If you want to get it merged, get a review from Redecorating

It looks like this change breaks file comparison and sticks new file in the tar archive. I dont think this will break anything but I dont want to take chances. It significantly improves performance for the python part (~0.3 seconds), so im going to try to replace sha256 with something faster since we dont care about security

AdityaGarg8:
Has checking for missing dependencies log in method 2 and 3 been removed?

AdityaGarg8:
> > I would avoid changes to python code. If you want to get it merged, get a review from Redecorating
> 
> It looks like this change breaks file comparison and sticks new file in the tar archive. I dont think this will break anything but I dont want to take chances. It significantly improves performance for the python part (~0.3 seconds), so im going to try to replace sha256 with something faster since we dont care about security

Sorry I won't merge any change to the python code

sharpenedblade:
> > > I would avoid changes to python code. If you want to get it merged, get a review from Redecorating
> > 
> > 
> > It looks like this change breaks file comparison and sticks new file in the tar archive. I dont think this will break anything but I dont want to take chances. It significantly improves performance for the python part (~0.3 seconds), so im going to try to replace sha256 with something faster since we dont care about security
> 
> Sorry I won't merge any change to the python code

Il open a different PR with it then, but with some work it can be significantly sped up.

sharpenedblade:
> Has checking for missing dependencies log in method 2 and 3 been removed?

No

AdityaGarg8:
> > > > I would avoid changes to python code. If you want to get it merged, get a review from Redecorating
> > > 
> > > 
> > > It looks like this change breaks file comparison and sticks new file in the tar archive. I dont think this will break anything but I dont want to take chances. It significantly improves performance for the python part (~0.3 seconds), so im going to try to replace sha256 with something faster since we dont care about security
> > 
> > 
> > Sorry I won't merge any change to the python code
> 
> Il open a different PR with it then, but with some work it can be significantly sped up.

As said before, if you want to make changes to python code, unless it's very necessary, first open it upstream.

sharpenedblade:
> > > > > I would avoid changes to python code. If you want to get it merged, get a review from Redecorating
> > > > 
> > > > 
> > > > It looks like this change breaks file comparison and sticks new file in the tar archive. I dont think this will break anything but I dont want to take chances. It significantly improves performance for the python part (~0.3 seconds), so im going to try to replace sha256 with something faster since we dont care about security
> > > 
> > > 
> > > Sorry I won't merge any change to the python code
> > 
> > 
> > Il open a different PR with it then, but with some work it can be significantly sped up.
> 
> As said before, if you want to make changes to python code, unless it's very necessary, first open it upstream.

I dont think upstream will accept a non-cryptographic hash, but since we arent verifying hashes anyways it doesnt matter to us

AdityaGarg8:
> > Has checking for missing dependencies log in method 2 and 3 been removed?
> 
> No, it got moved to the function that needs the dependency though

I can't find in the macOS bit

sharpenedblade:
> > > Has checking for missing dependencies log in method 2 and 3 been removed?
> > 
> > 
> > No, it got moved to the function that needs the dependency though
> 
> I can't find in the macOS bit

The python check function was removed and the code was inlined. Look at line 460.

sharpenedblade:
I tested the script it looks like its working now.

AdityaGarg8:
You have added checking for dependencies from python check. It's not there in deb/rpm/Pacman one

sharpenedblade:
> You have added checking for dependencies from python check. It's not there in deb/rpm/Pacman one

Then it says `Checking for dependencies` twice, which looks kinda wierd. Should I still add it.

AdityaGarg8:
> > You have added checking for dependencies from python check. It's not there in deb/rpm/Pacman one
> 
> Then it says `Checking for dependencies` twice, which looks kinda wierd. Should I still add it.

You should add it before the code that is executed first

sharpenedblade:
Something happened to the apfs method its really broken. Il mark this PR as ready once I sort out the bugs.

AdityaGarg8:
image
This is also broken

AdityaGarg8:
Tbh I am not really convinced in moving the non repetitive code like method 4 and 5 in functions. The code is much easier to read without them. So I would prefer reverting those changes.

AdityaGarg8 (changes requested):
Keep the code for APFS and macOS Recovery as it is.

sharpenedblade:
> Tbh I am not really convinced in moving the non repetitive code like method 4 and 5 in functions. The code is much easier to read without them. So I would prefer reverting those changes.

I reverted those.

sharpenedblade:
Everything is working now. I think the logs are fixed too.

docs/tools/firmware.sh
```diff
@@ -5,500 +5,789 @@
...
+				sudo modprobe -r brcmfmac
+				sudo modprobe brcmfmac
+				sudo modprobe -r hci_bcm4377
+				sudo modprobe hci_bcm4377
+				EOF
 				;;
 			(3)
-				echo -e "\nWhat package manager does your Linux distribution use?"
-				echo -e "\n1. apt"
+				echo -e "\nWhat package manager does your Linux distribution use?\n"
```
AdityaGarg8, line 778: You probably forgot to use cat <<-EOF here

AdityaGarg8 (changes requested):

docs/tools/firmware.sh
sharpenedblade: I left it like that on purpose, echo looks more like the final output. I can change it if you want though

AdityaGarg8: It's fine that way. I just said cause you were using cat so I thought you missed it by mistake.

docs/tools/firmware.sh
```diff
@@ -658,27 +937,23 @@ case "$os" in
...
 				python3 fetch-macOS-v2.py
 				echo -e "\nConverting image from .dmg to .img"
-				if [[ ${verbose} = -v ]]
+				if [[ ${verbose} = "-v" ]]
 				then
 					dmg2img -v BaseSystem.dmg fw.img
 				else
 					dmg2img -s BaseSystem.dmg fw.img
 				fi
+				rm BaseSystem.dmg
```
AdityaGarg8, line 946: rm ${verbose}

docs/tools/firmware.sh
```diff
@@ -5,500 +5,789 @@
...
+	echo -e "$package is missing!\n"
+	read -rp "Press enter to install $package and its dependencies. Alternatively you can terminate this script by pressing Control+C and install $package yourself, if you want to install it via some alternate method."
+
+	case $package_manager in
+		"apt")
+			case $package in
+				"linux-apfs-rw")
+					sudo apt upgrade
+					sudo apt install --reinstall -y apfs-dkms
+					sudo modprobe apfs
```
AdityaGarg8, line 409: modprobe ${verbose}

docs/tools/firmware.sh
```diff
@@ -5,500 +5,789 @@
...
+					sudo apt install -y "$package" ;;
+			esac ;;
+		"dnf")
+			case $package in
+				"linux-apfs-rw")
+					sudo dnf -y copr enable sharpenedblade/t2linux
+					sudo dnf install -y linux-apfs-rw
+					echo -e "\nRunning akmods\n"
+					sudo akmods
+					sudo modprobe apfs
```
AdityaGarg8, line 422: modprobe ${verbose}

docs/tools/firmware.sh
```diff
@@ -5,500 +5,789 @@
...
-				(pacman)
-					sudo pacman -Sy --noconfirm curl
+				*)
+					sudo dnf install -y "$package" ;;
+			esac ;;
+		"pacman")
+			case $package in
+				"linux-apfs-rw")
+					aur_install linux-apfs-rw-dkms-git
+					sudo modprobe apfs
```
AdityaGarg8, line 431: modprobe ${verbose}

AdityaGarg8 (changes requested):

docs/tools/firmware.sh
```diff
@@ -517,101 +806,96 @@ case "$os" in
...
+					sudo rm -r ${verbose} "${workdir}" || true
 					for i in 0 1 2 3 4 5
 					do
-						if [[ ${verbose} = -v ]]
-						then
-							sudo umount -v ${macosdir}/vol${i} || true
-						else
-							sudo umount ${macosdir}/vol${i} 2>/dev/null || true
-						fi
+						sudo umount "${macosdir}/vol${i}" 2>&1 | log || true
```
AdityaGarg8, line 876: Will this change mimic umount -v if run with -v

AdityaGarg8 (changes requested):

docs/tools/firmware.sh
```diff
@@ -5,500 +5,789 @@
...
-					cd - >/dev/null
-				fi
-				echo -e "\nFirmware tarball saved to Downloads!"
-				echo -e "\nExtract the tarball contents to /lib/firmware/brcm in Linux and run the following in the Linux terminal:"
-				echo -e "\nsudo modprobe -r brcmfmac_wcc"
-				echo "sudo modprobe -r brcmfmac"
-				echo "sudo modprobe brcmfmac"
-				echo "sudo modprobe -r hci_bcm4377"
-				echo "sudo modprobe hci_bcm4377"
+				create_firmware_archive /usr/share/firmware "$HOME/Downloads/firmware.tar" ${verbose}
```
AdityaGarg8, line 764: You have already specified verbose in the create_firmware_archive function. Is it required here as well?

docs/tools/firmware.sh
```diff
@@ -517,101 +806,96 @@ case "$os" in
...
 				mountpoint=$(mktemp -d)
 				workdir=$(mktemp -d)
 				echo "Installing Wi-Fi and Bluetooth firmware"
-				sudo mount ${verbose} /dev/nvme0n1p1 $mountpoint
-				sudo tar --warning=no-unknown-keyword ${verbose} -xC ${workdir} -f $mountpoint/firmware.tar.gz
-				sudo python3 "$0" ${workdir} ${workdir}/firmware-renamed.tar ${verbose}
+				sudo mount ${verbose} /dev/nvme0n1p1 "$mountpoint"
+				sudo tar --warning=no-unknown-keyword ${verbose} -xC "${workdir}" -f "$mountpoint/firmware-raw.tar.gz"
+				sudo chown -R "$USER" "${workdir}"
+				create_firmware_archive "${workdir}" "${workdir}/firmware-renamed.tar" ${verbose}
```
AdityaGarg8, line 837: Here also you have added verbose

docs/tools/firmware.sh
```diff
@@ -621,35 +905,30 @@ case "$os" in
...
 				fi
-				python3 "$0" ${fwlocation} ${workdir}/firmware-renamed.tar ${verbose} || (echo -e "\nCouldn't extract firmware. Try running the script again. If error still persists, try restarting your Mac and then run the script again, or choose some other method." && unmount_macos_and_cleanup && exit 1)
-				sudo tar ${verbose} -xC /lib/firmware/brcm -f ${workdir}/firmware-renamed.tar
-				echo "Reloading Wi-Fi and Bluetooth drivers"
-				sudo modprobe -r ${verbose} brcmfmac_wcc || true
-				sudo modprobe -r ${verbose} brcmfmac || true
-				sudo modprobe ${verbose} brcmfmac || true
-				sudo modprobe -r ${verbose} hci_bcm4377 || true
-				sudo modprobe ${verbose} hci_bcm4377 || true
+				create_firmware_archive "${fwlocation}" "${workdir}/firmware-renamed.tar" "${verbose}" || (echo -e "\nCouldn't extract firmware. Try running the script again. If error still persists, try restarting your Mac and then run the script again, or choose some other method." && unmount_macos_and_cleanup && exit 1)
```
AdityaGarg8, line 908: Here also verbose

docs/tools/firmware.sh
```diff
@@ -658,27 +937,23 @@ case "$os" in
...
 				cd - >/dev/null
-				python3 "$0" ${imgdir}/usr/share/firmware ${workdir}/firmware-renamed.tar ${verbose} || (echo -e "\nCouldn't extract firmware. Try choosing some other macOS version (should be Monterey or later). If error still persists, try restarting your Mac and then run the script again." && cleanup_dmg && exit 1)
-				sudo tar ${verbose} -xC /lib/firmware/brcm -f ${workdir}/firmware-renamed.tar
-				echo "Reloading Wi-Fi and Bluetooth drivers"
-				sudo modprobe -r ${verbose} brcmfmac_wcc || true
-				sudo modprobe -r ${verbose} brcmfmac || true
-				sudo modprobe ${verbose} brcmfmac || true
-				sudo modprobe -r ${verbose} hci_bcm4377 || true
-				sudo modprobe ${verbose} hci_bcm4377 || true
+				create_firmware_archive "${imgdir}/usr/share/firmware" "${workdir}/firmware-renamed.tar" ${verbose} || (echo -e "\nCouldn't extract firmware. Try choosing some other macOS version (should be Monterey or later). If error still persists, try restarting your Mac and then run the script again." && cleanup_dmg && exit 1)
```
AdityaGarg8, line 954: Here also verbose added

docs/tools/firmware.sh
```diff
@@ -517,101 +806,96 @@ case "$os" in
...
-							sudo rm ${verbose} $file
+							sudo rm ${verbose} "$file"
 						fi
 					done
 				fi
-				sudo rm -r ${verbose} ${workdir}
-				sudo umount $mountpoint
-				sudo rmdir $mountpoint
+				sudo rm -r ${verbose} "${workdir}"
+				sudo umount "$mountpoint"
```
AdityaGarg8, line 864: umount ${verbose}

docs/tools/firmware.sh
```diff
@@ -517,101 +806,96 @@ case "$os" in
...
+				sudo rmdir "$mountpoint"
```
AdityaGarg8, line 865: rmdir ${verbose}

AdityaGarg8 (changes requested):

AdityaGarg8:
Don't squash the commit addressing these changes.

docs/tools/firmware.sh
AdityaGarg8: Btw in cleanup_dmg function, workdir gets removed, which also has the dmg, so I wonder what's the actual advantage of specifically removing it here, since the next step before cleanup is simply firmware extraction and reloading drivers.

sharpenedblade: tmpfs is usually in ram, if you are running low on ram the script crashes. The DMG is almost 2GB so this saves a lot of ram space.

docs/tools/firmware.sh
sharpenedblade: I need to pass`${verbose}` to umount. the `| log` just ignores whatever is passed to it in normal mode and outputs to stdout in verbose mode.

docs/tools/firmware.sh
sharpenedblade: Its a leftover from when I had `verbose=$3`, it doesnt do anything now. Im going to remove it

docs/tools/firmware.sh
sharpenedblade: Same as above

docs/tools/firmware.sh
AdityaGarg8: > tmpfs is usually in ram, if you are running low on ram the script crashes. The DMG is almost 2GB so this saves a lot of ram space.

But it is not removed unless img is made, so what's the benefit here?

docs/tools/firmware.sh
AdityaGarg8: Ok

docs/tools/firmware.sh
sharpenedblade: When I was testing the script it would crash a lot here and complain about low disk space, but once I added the `rm` it usually worked.

sharpenedblade: Im going to move the dmg dir to somewhere on disk.

AdityaGarg8:
nvram_txcap_quirk is broken in Method 1 (I haven't tested other methods yet). Cause is quite evident by the code itself. You are making a firmware-raw.tar.gz for the main firmware and firmware-raw.tar for extra firmware. After that you are only extracting firmware-raw.tar.gz on Linux and firmware-raw.tar is untouched.

Screenshot 2024-06-18 at 10 43 25 PM

I don’t understand how to moved the extra files to workdir to Linux as well.

**Please test whatever you change.**

AdityaGarg8:
Also, separate out python_check for macOS in a separate function like it was in the older version. There is not need to unnecessarily complicate the create_firmware_archive

sharpenedblade:
> I don’t understand how you moved the extra files to workdir on Linux as well.

Once I add those files to the right archive it will work. When the tarfile is extracted those files will be in $PWD, which is $workdir

> **Please test whatever you change.**

I dont have a mac where that code is used.

> Also, separate out python_check for macOS in a separate function like it was in the older version. There is not need to unnecessarily complicate the create_firmware_archive

Ok

AdityaGarg8:
> > I don’t understand how you moved the extra files to workdir on Linux as well.
> 
> Once I add those files to the right archive it will work. When the tarfile is extracted those files will be in $PWD, which is $workdir
> 
> > **Please test whatever you change.**
> 
> I dont have a mac where that code is used.

You can change the model identifier to your Mac and path to the firmware folder to your chip. That's how I test on my Mac.
> 
> > Also, separate out python_check for macOS in a separate function like it was in the older version. There is not need to unnecessarily complicate the create_firmware_archive
> 
> Ok

AdityaGarg8:
What are you adding python_check in create_firmware_archive? Just put it in macOS specific code.

sharpenedblade:
> What are you adding python_check in create_firmware_archive? Just put it in macOS specific code.

I moved it

AdityaGarg8:
Script looks good to me now, I've tested it on Ubuntu as well as Arch. I wonder if there could be a way to use /tmp if storage in RAM is sufficient. I would prefer not to write much on the SSD for such tasks.

sharpenedblade:
> I wonder if there could be a way to use /tmp if storage in RAM is sufficient.

I think its better to use the disk since a lot of macs only have 8Gb of RAM and if you have swap enabled its going to write to disk anyways.

>  I would prefer not to write much on the SSD for such tasks.

It shouldn't be a problem, a lot of things write more than this to disk. SSD wear only matters if you write large amounts of data regularly, a user will probably only run this script once.

docs/tools/firmware.sh
```diff
@@ -621,35 +918,30 @@ case "$os" in
...
-					sudo losetup -d /dev/${loopdev}
+					sudo rm -r ${verbose} "${workdir}"
+					sudo umount ${verbose} "${loopdev_partition}"
+					sudo rm -r ${verbose} "${imgdir}"
+					sudo losetup -d /dev/"${loopdev}"
 				}
 
 				echo -e "\nDownloading macOS Recovery Image"
-				workdir=$(mktemp -d)
+				workdir=$(mktemp -d "$HOME/.tmp.XXXXXX")
```
AdityaGarg8, line 941: Don’t make it hidden. In case the script halts in between, such as user presses control+c, Mac discharges etc, the folder is visible to the user and he can remove it manually.

AdityaGarg8 (changes requested):

PR #554, opened 2024-06-29
Add json metadata for the ISO and distro list

sharpenedblade:
This is for a gui installer I am writing, I wanted a parsable list of ISOs and distros. It has every ISO from the preinstall guide except for gentoo and blendos. This needs to be updated whenever you release a new ISO.

`iso_compression` should be `zip` if the ISO file is the only file inside of a zip archive. If the ISO is *not* compressed then the `iso_compression` field should not appear (it should not be null or empty).
```
{
  "all": [
    {
      "name": "user visible disto name",
      "iso": ["url to iso download", "appended in order to the first file"],
      "iso_compression": "zip",
    }
  ]
}
```

### DIFF
--- a/docs/tools/distro-metadata.json
+++ b/docs/tools/distro-metadata.json
@@ -22,31 +22,31 @@
       "iso": ["https://github.com/t2linux/fedora-iso/releases/download/f40.0/t2linux-fedora-live-sway-40.iso"]
     },
     {
-      "name": "Ubuntu (Jammy)",
+      "name": "Ubuntu 22.04 - Jammy Jellyfish",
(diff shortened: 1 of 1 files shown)


AdityaGarg8 (changes requested):
I'd prefer keeping Ubuntu, Kubuntu and linux mint together, in this order.

Also prefer using heading: Ubuntu (22.04 Jammy Jellyfish)

sharpenedblade:
> I'd prefer keeping Ubuntu, Kubuntu and linux mint together, in this order.

I didnt put them in any sort of real order, I can move the other ones around too.

AdityaGarg8:
Can you verfiy https://github.com/t2linux/wiki/pull/554/commits/45857a3c6e9420319acf742ea66898337b0d91de (for symbols not allowed, other improvements etc)

sharpenedblade:
It looks good. Anything that's valid utf-8 is allowed inside of the name field, so you can even use emojis and stuff.

PR #573, opened 2024-09-22
Add missing OS spoof step in enabling integrated GPU

dullbananas:


AdityaGarg8:
I believe you are using an old Linux kernel, or haven't tested the guide yourself cause os spoof is now integrated into the Linux kernel.

PR #582, opened 2024-10-15
fix rpm creation in firmware.sh

adrianinsaval:
otherwise this fails because `create_package` expects dnf as value, not rpm


sharpenedblade:
Looks good to me, the rest of the code uses and expects `dnf`, im not sure why this was `rpm`.

sharpenedblade (approved):

PR #604, opened 2025-02-03
NixOS: fix declarative firmware setup code

ddogfoodd:
- the added line specifies to only run the installPhase when building the derivation.
- this is fine since we don't need other phases here
- this fixes an otherwise returned error about the unpackPhase not being defined


AdityaGarg8:
@soopyc can you review this please.

soopyc:
tentative lgtm, will report back in a bit. i will note that in my test builds (including one that i just did) i've never seen any errors or warnings about unpackPhase with that derivation.

soopyc:
On 4 February 2025 12:15:22 am HKT, Jost Alemann ***@***.***> wrote:
>Still, thanks for the review and double-check @soopyc ! Sorry for stealing your time as well
>

no worries, nix is flaky sometimes.

PR #609, opened 2025-02-08
Added iwd Backend Details

Alexb715:
I added details on how to add iwd to distro's (i was targeting arch cause that what i tested on) that don't automatically sets it as its backend


AdityaGarg8:
I need to add more details, your PR needs a rewrite. I'll add it myself later.

PR #610, opened 2025-02-08
Wifi Bluetooth Guide Update

xygamerz:
## Description
This PR replaces `wpa_supplicant` with `iwd` as the Wi-Fi backend due to an upstream issue with `wpa_supplicant`. The change ensures a more stable and reliable Wi-Fi connection for users.

## Changes 
Added a warning message in the documentation to alert users about the `wpa_supplicant `issue.
Recommended `iwd` as the preferred Wi-Fi backend until the issue is resolved.
## Why this change is necessary
There is a known upstream issue with `wpa_supplicant` that may cause instability or connectivity problems.
`iwd` is a modern and reliable alternative that can help users avoid these issues.
This warning ensures users are aware of the problem 
## Possible Addition
I could add a guide at the bottom of the page to advise user how to fix the issue if wanted


sharpenedblade:
Can you mention that this is not necessary on Fedora

xygamerz:
I add a note as wanted

sharpenedblade (approved):
It looks good to me

PR #613, opened 2025-02-15
docs: add cleanup commands

aloshy-ai:
- Add nix-editor command to remove templating configuration
- Add sed command to remove unnecessary comments in flake.nix

### DIFF
--- a/docs/distributions/nixos/installation.md
+++ b/docs/distributions/nixos/installation.md
@@ -104,11 +104,14 @@ You might want to also configure a display manager and a desktop environment. Ch
         #    Accept *both* settings if you would like to use the substituter.
         nix flake init -t github:soopyc/nixos-t2-flake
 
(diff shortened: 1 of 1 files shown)


AdityaGarg8:
Cc @soopyc

soopyc:
as mentioned in a different repo: https://github.com/soopyc/nixos-t2-flake/pull/1#discussion_r1957107872 this command only works on macos and will need updating, especially because this would usually be run in the live environment and not macos.

imo this should be left as an exercise to the user.

AdityaGarg8:
> as mentioned in a different repo: [soopyc/nixos-t2-flake#1 (comment)](https://github.com/soopyc/nixos-t2-flake/pull/1#discussion_r1957107872) this command only works on macos and will need updating, especially because this would usually be run in the live environment and not macos.
> 
> imo this should be left as an exercise to the user.

Even I would prefer to manually edit the file instead of using sed commands

PR #615, opened 2025-02-21
Disable offloading to fix WiFi issue on Broadcom chips

graemer957:
I tried switching to `iwd`, but it did not fix the problem for me. This [thread](https://bbs.archlinux.org/viewtopic.php?id=298025) contained the answer... disabling offloading.

### DIFF
--- a/docs/guides/wifi-bluetooth.md
+++ b/docs/guides/wifi-bluetooth.md
@@ -5,7 +5,7 @@
 This page is a step by step guide to get Wi-Fi and Bluetooth working on T2 Macs. This guide is also applicable to **iMac19,1** and **iMac19,2**, which are T1 Intel Macs. This guide is NOT applicable for the rest of the T1 and older Intel Macs.
 
 !!! Bug "Unable to connect to Wi-Fi even after entering correct password"
-    Due to a [regression](https://lists.infradead.org/pipermail/hostap/2024-August/042893.html) since **wpa_supplicant 2.11**, Wi-Fi on broadcom chips is broken, thus affecting the T2 Macs as well. Please use [iwd](https://wiki.archlinux.org/title/NetworkManager#Using_iwd_as_the_Wi-Fi_backend) or [disable offloading](https://bbs.archlinux.org/viewtopic.php?pid=2189161#p2189161) until the issue is resolved. Currently, only Arch Linux and EndeavourOS are affected.
+    Due to a [regression](https://lists.infradead.org/pipermail/hostap/2024-August/042893.html) since **wpa_supplicant 2.11**, Wi-Fi on broadcom chips is broken, thus affecting the T2 Macs as well. Please use [iwd](https://wiki.archlinux.org/title/NetworkManager#Using_iwd_as_the_Wi-Fi_backend) or [disable offloading](https://bbs.archlinux.org/viewtopic.php?pid=2189161#p2189161) by adding `brcmfmac.feature_disable=0x82000` to your kernel parameters until the issue is resolved. Currently, only Arch Linux and EndeavourOS are affected.
 
 !!! Warning "Arch/EndeavourOS"
     If you're running Arch or EndeavourOS and have `apple-bcm-firmware` installed, you do not need to follow this guide further.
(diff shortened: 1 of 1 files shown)


sharpenedblade:
Switching to `iwd` should work since the bug is only in wpa_supplicant >= 2.11. Are you sure you properly switched to iwd?

AdityaGarg8:
Probably just say:

"disable offloading by adding `brcmfmac.feature_disable=0x82000` to your kernel parameters"

AdityaGarg8:
> I believe so, I:
> 
> 1. Installed `iwd`
> 2. Disabled `wpa_supplicant`
> 3. Added a `.conf` file with:
>    ```
>    [device]
>    wifi.backend=iwd
>    ```

What is the path of the conf file
> 
> Trying to connect to WiFi would just result in repeated entries from `iwd`:

Did you restart Network manager after adding the file?
> 
> ```
> event: connect-failed, status: 16
> ```

graemer957:
> What is the path of the conf file

`/etc/NetworkManager/conf.d`

> Did you restart Network manager after adding the file?

Yes

> Also, what distro are you using?

Arch

graemer957:
> Probably just say:
> 
> "disable offloading by adding `brcmfmac.feature_disable=0x82000` to your kernel parameters"

I would be happy to update accordingly, once we have reached quorum on evening making an addition 😄

AdityaGarg8:
> > Probably just say:
> > "disable offloading by adding `brcmfmac.feature_disable=0x82000` to your kernel parameters"
> 
> I would be happy to update accordingly, once we have reached quorum on evening making an addition 😄

We can add this. It's equally good.

AdityaGarg8:
Can you test this btw. I think you made some mistake while switching to iwd before.

```bash
cat <<EOF | sudo tee /etc/NetworkManager/conf.d/wifi_backend.conf
[device]
wifi.backend=iwd
EOF
sudo systemctl restart NetworkManager
```

docs/guides/wifi-bluetooth.md
```diff
@@ -5,7 +5,7 @@
...
 
 !!! Bug "Unable to connect to Wi-Fi even after entering correct password"
-    Due to a [regression](https://lists.infradead.org/pipermail/hostap/2024-August/042893.html) since **wpa_supplicant 2.11**, Wi-Fi on broadcom chips is broken, thus affecting the T2 Macs as well. Please use [iwd](https://wiki.archlinux.org/title/NetworkManager#Using_iwd_as_the_Wi-Fi_backend) or [disable offloading](https://bbs.archlinux.org/viewtopic.php?pid=2189161#p2189161) until the issue is resolved. Currently, only Arch Linux and EndeavourOS are affected.
+    Due to a [regression](https://lists.infradead.org/pipermail/hostap/2024-August/042893.html) since **wpa_supplicant 2.11**, Wi-Fi on broadcom chips is broken, thus affecting the T2 Macs as well. Please use [iwd](https://wiki.archlinux.org/title/NetworkManager#Using_iwd_as_the_Wi-Fi_backend) or [disable offloading](https://bbs.archlinux.org/viewtopic.php?pid=2189161#p2189161) by adding `brcmfmac.feature_disable=0x82000` to your kernel parameters until the issue is resolved. Currently, only Arch Linux and EndeavourOS are affected.
```
AdityaGarg8, line 8: Just remove the link. This link is more confusing that useful. Instead make **disable offloading** bold.

AdityaGarg8 (changes requested):

docs/guides/wifi-bluetooth.md
AdityaGarg8: Also make it 

"please use iwd as a backend or disable offloading...."

graemer957:
> Can you test this btw. I think you made some mistake while switching to iwd before.

This does not work for me. The fix for me has been to disable offloading.

docs/guides/wifi-bluetooth.md
graemer957: I suggest making suggestions using the provided mechanism in GitHub to avoid any mistakes.

PR #617, opened 2025-03-08
Update fedora installation guide

sharpenedblade:


docs/distributions/fedora/installation.md
```diff
@@ -21,6 +21,9 @@ If you need a more updated kernel, use the iso from [sharpenedblade](https://git
 
 ## Installing unsupported spins
 
-1. Follow the installation instructions above, but use your custom (vanilla) ISO. You need an external keyboard and mouse.
+1. Follow the installation instructions above, but use your custom (vanilla) ISO. You need an external keyboard and mouse. WiFi will not work until you finish the installation procedure so you need a wired internet connection.
```
AdityaGarg8, line 24: Cannot we use the wifi script to dump firmware in the ISO?

sharpenedblade: The wifi script is in the `t2linux-scripts` package so you have to manually copy it onto the ISO if you want it.

AdityaGarg8: > The wifi script is in the `t2linux-scripts` package so you have to manually copy it onto the ISO if you want it. Also does wifi work with the upstream kernel?

WiFi works with upstream kernel. Support has been merged long ago.

AdityaGarg8: We can modify this later then, I'll merge this for now.

PR #624, opened 2025-04-17
Update get-apple-firmware.service activation logic

sharpenedblade:
This creates the file /etc/get_apple_firmware_attempted when the service is activated. The service will only activate when this file is *not* present. The service will run once on first boot and will disable itself after, regardless of the exit status of the firmware script. This change should make the service actually start on most distros since many do not follow systemd first-boot conventions.

### DIFF
--- a/docs/tools/get-apple-firmware.service
+++ b/docs/tools/get-apple-firmware.service
@@ -1,14 +1,13 @@
 [Unit]
 Description=Get Apple WiFi and Bluetooth firmware
-ConditionFirstBoot=yes
+ConditionPathExists=!/etc/get_apple_firmware_attempted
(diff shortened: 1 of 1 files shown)


sharpenedblade:
Should I change the service to extract from macos instead of the efi tarball by default?

AdityaGarg8:
> Should I change the service to extract from macos instead of the efi tarball by default?

Isn't it extracting from macOS currently? Also I forgot what was -i for.

sharpenedblade:
> Isn't it extracting from macOS currently?

Oh sorry, I have a unpushed local commit so it looked like it was getting it from efi for me.


> Also I forgot what was -i for.

It disables interactive mode, so it won't prompt for user input and will just exit.

PR #688, opened 2026-02-20
Add workaround for BCM4377b Wi-Fi failing after lid close

herz-j:
Adds documentation for a known issue on T2 Macs with the BCM4377b Wi-Fi chip (confirmed on MacBookAir9,1) where brcmfmac fails to enter D3 power state during suspend, leaving Wi-Fi and Bluetooth unresponsive after the lid is opened.

The root cause is a Broadcom firmware bug that has been reported upstream but has no fix available. The workaround involves forcing s2idle sleep and using a post-resume systemd script to rebind the driver and reload the firmware.

### DIFF
--- a/docs/guides/wifi-bluetooth.md
+++ b/docs/guides/wifi-bluetooth.md
@@ -237,3 +237,63 @@ Dec 24 22:34:20 hostname kernel: brcmfmac: brcmf_c_process_txcap_blob: TxCap blo
 Dec 24 22:34:20 hostname kernel: brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4377/4 wl0: Jul 16 2021 18:25:13 version 16.20.328.0.3.6.105 FWID 01-30be2b3a
 Dec 24 22:34:20 hostname kernel: brcmfmac 0000:01:00.0 wlp1s0f0: renamed from wlan0
 ```
+## Wi-Fi not working after lid close (BCM4377b)
+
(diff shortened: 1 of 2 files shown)


docs/guides/wifi-bluetooth.md
```diff
@@ -237,3 +237,63 @@ Dec 24 22:34:20 hostname kernel: brcmfmac: brcmf_c_process_txcap_blob: TxCap blo
 Dec 24 22:34:20 hostname kernel: brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4377/4 wl0: Jul 16 2021 18:25:13 version 16.20.328.0.3.6.105 FWID 01-30be2b3a
 Dec 24 22:34:20 hostname kernel: brcmfmac 0000:01:00.0 wlp1s0f0: renamed from wlan0
 ```
+## Wi-Fi not working after lid close (BCM4377b)
+
+!!! Bug "Affected hardware: MacBookAir9,1 and possibly other T2 Macs with BCM4377b"
+    On some T2 Macs, the `brcmfmac` driver fails to enter D3 power state during suspend,
+    causing the Wi-Fi firmware to become unresponsive after the lid is opened. This is a
+    known Broadcom firmware bug [reported to Broadcom](https://github.com/t2linux/T2-Debian-and-Ubuntu-Kernel/issues)
+    with no upstream fix available. The workaround below resolves it.
```
sharpenedblade: I think you used the wrong link here

PR #697, opened 2026-03-20
docs: add kexec reboot workaround for T2 Macs

Asrafiel:
T2 Macs power off instead of rebooting when reboot is called. The T2 firmware ignores the ACPI reboot command. This documents a kexec-based workaround that bypasses firmware by loading the kernel into memory before shutdown sequence unmounts filesystems.

Critical detail: the systemd service must run Before=umount.target or /boot is gone before the script can find the kernel.


kevin-david:
@Asrafiel I'm curious if this was a consistent or intermittent problem for you? I had the same thing happen today, once, but then the next reboot was fine ...

deqrocks:
This was just the t2 panicking because old apple-bce didn't have pm paths for shutdown/reboot. So yes. AI slop

PR #713, opened 2026-05-12
Add note for non-T2 MacBook Pro A1708

alioalyasari:
I know my machine isn’t t2 but this guide helped me a lot so trying to give back. Not sure if this is the right place I’m new to Linux and GitHub.


AdityaGarg8:
Thanks for the contribution but I don't find the wiki as the right place for this guide.

PR #720, opened 2026-06-15
Link rEFInd boot settings from the audio guide

igorsantos07:
Also:
- move back from now-erased t2-better-audio folder in the suggested bash script
- fix title formatting to comply with the others

------

In my case, the original audio guide suggesting to edit GRUB stuff didn't help with `/proc/cmdline`, but after creating `refind_linux.conf` and adding the proper options there, it did the trick.
That doesn't mean I got audio to work, but at least the correct boot options are now showing up :joy:


AdityaGarg8:
Please fix the lint errors

igorsantos07:
Sorry. Done!

PR #723, opened 2026-06-21
clarifies force igd and adds dgpu disable

deqrocks:


AdityaGarg8:
Vgaswitcheroo doesn't work yet with MacBook Pro 16,1

deqrocks:
> Vgaswitcheroo doesn't work yet with MacBook Pro 16,1

It should for switching the dGPU off and on on boot. But not for runtime switching if that is what you mean. We don't use runtime switching here.

AdityaGarg8:
@Redecorating could you review this please?

Redecorating (approved):
lgtm

Redecorating:
switching the amdgpu off on mbp16,1 has been fine for me. its switching it back on that never worked.

AdityaGarg8:
> lgtm

Thank you!

Issue #543, opened 2024-05-29
The `Hybrid Graphics` guide results in a black screen on MBP16,1


Kir-Antipov:
Hi! First of all, thanks a lot for this project; it's been amazingly helpful to get a working keyboard, Wi-Fi, and other somewhat important stuff out of the box :D

However, I can't seem to get hybrid graphics working as described in the corresponding guide, which is especially strange considering it explicitly states that it should work on the MBP16,1. Whenever I set the `apple_set_os` kernel parameter via my GRUB config and run `echo "options apple-gmux force_igd=y" | sudo tee /etc/modprobe.d/apple-gmux.conf`, I just get a black screen on the subsequent reboot. *(I can see the logs running for a second or so, then I assume the switch to iGPU happens, and the screen goes dark.)* The only way to get out of this is to forcefully shut down the machine and then reboot it, omitting the `apple_set_os` parameter. Interestingly enough, `journalctl -b -1` suggests there were no graphics-related errors, and the system successfully reached the login screen before I pressed the power button. I understand that this whole issue *originally* stems from Apple firmware shutting down the iGPU if you try to boot something other than macOS *(wow, thanks Apple, very helpful!)*. However, to be quite frankly honest with you, I have never encountered problems related to powered-down graphics cards, so I'm not sure if this is what's happening in this exact case, as I would expect at least some graphics errors to show up if the system tries to use a disabled GPU.

If I remove either `apple_set_os` or `/etc/modprobe.d/apple-gmux.conf`, the device boots fine. In the first case, the iGPU is not recognized at all *(i.e., it's not listed via `lspci`)*, and in the latter, it is recognized, but the dGPU is used by default. So, at least this part seems to work as expected.

Also, I tried to effectively replace `apple-gmux.conf` with the `modprobe.blacklist=amdgpu` kernel parameter *(I don't really need hybrid graphics in the end of this journey, I just would love to bring the power consumption of the MBP down as much as possible)*, hoping it would result in a switch to the iGPU because the dGPU is simply not available. This does happen *(at least according to `glxinfo`)*, but this approach produces some weird graphical artifacts *(e.g., the cursor leaves a Windows 97-ish trail behind itself and other fun stuff)*, and the power consumption is still really high, signifying that the dGPU is not properly shut down, which is honestly to be expected with such a barbaric approach.

[...]

PR #738, opened 2026-07-28
Warn MacBookPro15,3 users about iGPU black screen

kevin-david:
## What changed
I tested the iGPU steps for my 2019 15" MacBook Pro and they did not work. Without remote access already set up this would have created a large annoyance for me, so... 

- Remove the unverified claim that MacBookPro15,3 should support the iGPU setup
- Document the black-screen failure observed on T2 kernel `7.1.3-1-t2-resolute`
- Include the relevant i915 VBT/CRTC errors
- Require a working Intel framebuffer before disabling the dGPU and document recovery

### DIFF
--- a/docs/guides/hybrid-graphics.md
+++ b/docs/guides/hybrid-graphics.md
@@ -2,7 +2,7 @@
 
 This page describes how to use the iGPU on MacBook Pros with Hybrid Graphics (2 GPUs). 13-inch MacBooks only have an iGPU and do not need this. Using the iGPU means you can save power by putting the more powerful AMD dGPU in a low power state when you don't need it.
 
-This has been tested on the MacBookPro16,1 and the MacBookPro15,1. The 15,3 and 16,4 models are very similar and should work too.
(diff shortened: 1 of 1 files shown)


kevin-david:
Actually, I'm not sure this is true yet, based on #543, closing for now.

kevin-david:
In case anyone has this issue, I resolved it by upgrading the T2 Linux kernel from 7.1.3-1-t2-resolute to 7.1.5-1-t2-resolute, and changed `gpu-power-prefs` to select the iGPU

```
gpu-power-prefs: 07 00 00 00 01 00 00 00
apple-gmux: force_igd=y
amdgpu-off.service: enabled
```

On 7.1.3, apple-gmux force_igd=y switched to Intel, but i915 couldn’t create a framebuffer allegedly because the VBT was missing. On 7.1.5, the VBT is still missing, but i915 successfully creates i915drmfb and drives the panel anyway. `/proc/fb` contained 0 i915drmfb instead of just IGD:+:Pwr

PR #754, opened 2026-09-06
Documentation on how to install without a USB drive

bartbroere:
Recently I managed to install Ubuntu on a 2018 Intel Macbook from a partition on the Mac's internal drive (instead of from an external USB drive).

This pull request adds documentation on how to do this. The reason it is still marked draft is that I would like to hear the experiences from other people. Was my successful Ubuntu installation a stroke of luck or would this work for all distributions? A simple "worked for me using Fedora" comment for example could already be helpful.

The motivation for this documentation is to help people avoid buying something new (either a USB-C storage device or adapter to USB-A). I suspect people installing Linux on an eight-year old Macbook generally don't like to solve problems by buying something. Even if you *have* the right dongle looking for it through the drawers might be a hassle slowing your installation down.

Now that 
+ [new versions of MacOS no longer support Intel Macbooks](https://en.wikipedia.org/wiki/Mac_transition_to_Apple_silicon#2025)
+ [Homebrew stopped supporting Intel based MacOS as of this month](https://docs.brew.sh/Support-Tiers#future-macos-support)
+ [Apple hardware support now is very complete in the Linux kernel](https://www.phoronix.com/news/Apple-Touch-Bar-Display-Linux)

I believe the T2 Linux project might get a lot of attention. The easier the installation process is for these new users, the better!


sharpenedblade:
You can also boot the ISO in a vm on macos, and pass the internal disk as a block device. Ive done this to install fedora and nixos with qemu and its pretty easy and doesn't make you mess around with partitions.

bartbroere:
> You can also boot the ISO in a vm on macos, and pass the internal disk as a block device. Ive done this to install fedora and nixos with qemu and its pretty easy and doesn't make you mess around with partitions.

That sounds like a very nice approach! I hope to try that soon and maybe even document the steps required.

Although it definitely avoids messing around with partitions, some issues might arise from the emulation done by QEMU. At first Linux boot you might have to do some setting up of hardware that wasn't emulated during the installation (wifi hardware maybe). How much work this is probably differs per distribution, but I'll try to find out.

To make it beginner friendly documenting this approach should start at installing brew, followed by using brew to install qemu.

sharpenedblade:
> Although it definitely avoids messing around with partitions, some issues might arise from the emulation done by QEMU. At first Linux boot you might have to do some setting up of hardware that wasn't emulated during the installation (wifi hardware maybe). How much work this is probably differs per distribution, but I'll try to find out.

It should just work on most distributions. It has for me.

> To make it beginner friendly documenting this approach should start at installing brew, followed by using brew to install qemu.

sound good to me

